PR #15, opened 2025-10-17
Publish scripts to run benchmarks

edeandrea:
Publishing all the scripts & documentation to run official benchmarks.

### DIFF
--- a/scripts/perf-lab/README.md
+++ b/scripts/perf-lab/README.md
@@ -60,6 +60,7 @@ The script also has 3 dependencies that need to be resolved before it can be run
 | `-c`   | `<CGROUPS_CPUS>`                 | Constrain to certain CPUs via [cgroups](https://en.wikipedia.org/wiki/Cgroups) (comma-separated list, e.g., `0,2,4,6,8,10,12,14`). Ignored if running on a host that doesn't support cgroups. |                                                                   |
 | `-d`   |                                  | Purge/drop OS filesystem caches between iterations                                                                                                                                            |                                                                   |
 | `-e`   | <EXTRA_QDUP_ARGS>                | Any extra arguments that need to be passed to qDup ahead of the qDup scripts.<br/> - **NOTE:** This is an advanced option. Make sure you know what you are doing when using it.               |                                                                   |
+| `-f`   | <OUTPUT_DIR>                     | The directory containing the run output                                                                                                                                                       | `/tmp`                                                            |
 | `-g`   | `<GRAALVM_VERSION>`              | The GraalVM version to use for native tests (from SDKMAN)                                                                                                                                     | `25-graalce`                                                      |
 | `-h`   | `<HOST>`                         | The HOST to run benchmarks on. Use `LOCAL` to run on the local machine                                                                                                                        | `LOCAL`                                                           |
 | `-i`   | `<ITERATIONS>`                   | The number of iterations to run each test. The output is the average across all iterations.                                                                                                   | `3`                                                               |
@@ -117,7 +118,6 @@ Finished in 09:44.800 at /tmp/20251021_090429
 If you examine the output directory:
 
 ```shell
-/tmp/20251021_090429
 ├── run.json
 ├── run.log
 └── target-host


edeandrea:
@holly-cummins @franz1981 @willr3 would you all mind reviewing?

I still need to add documentation

benchmarks/run-benchmarks.sh
```diff
@@ -0,0 +1,265 @@
...
+  echo "  -n <NATIVE_QUARKUS_BUILD_OPTIONS>   Native build options to be passed to Quarkus native build process"
+  echo "  -o <NATIVE_SPRING_BUILD_OPTIONS>    Native build options to be passed to Spring native build process"
+  echo "  -p <PROFILER>                       Enable profiling with async profiler"
+  echo "                                          Accepted values: none, jfr, flamegraph"
+  echo "                                          Default: none"
+  echo "  -q <QUARKUS_VERSION>                The Quarkus version to use"
+  echo "  -r <RUNTIMES>                       The runtimes to test"
+  echo "                                          Accepted values (1 or more of): quarkus3-jvm, quarkus3-native, spring3-jvm, spring3-jvm-aot, spring3-native"
+  echo "                                          Default: 'quarkus3-jvm,quarkus3-native,spring3-jvm,spring3-jvm-aot,spring3-native'"
+  echo "  -s <SPRING_BOOT_VERSION>            The Spring Boot version to use"
```
franz1981: Ditto as https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/15/files#r2443753270

Not specifying both is clearly not cool, since it means nothing has to run.
But generally speaking it makes super simple to try the script even without any comparison in mind e.g. 

```bash
./run-benchmarks.sh  -q 3.28.4
```
will just work as it is

benchmarks/main.yml
```diff
@@ -0,0 +1,437 @@
...
+  config.cgroup.mem_max: "14G"
+  config.cgroup.name: spring-quarkus-perf-comparison
+
+  config.profiler.name: none #jfr flamegraph
+  config.profiler.events: cpu
+
+  config.repo.branch: main
+  config.repo.url: https://github.com/quarkusio/spring-quarkus-perf-comparison.git
+
+  RUN.WRK_BIN: jbang wrk@hyperfoil
```
franz1981: This is what is reported by `wrk`  while extracting manually the report:

[wrk.html](https://github.com/user-attachments/files/22995209/wrk.html)

Some notes:
1. calibration phase duration is hardcoded in the command (which is not great)
2. the report is pretty boring per se: latencies are useless (due to coordinated omission) and, as expected, throughput is stable

Which makes the report per se not as interesting as I was hoping (due to the simplicity of the workload...)
In short: I don't think we need to extract the report if we want to keep this test as simple as this.

franz1981:
@edeandrea re https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/15#issuecomment-3420445120 I've commented at https://github.com/Hyperfoil/qDup/issues/250

benchmarks/run-benchmarks.sh
edeandrea: I've updated the script docs

benchmarks/main.yml
```diff
@@ -0,0 +1,437 @@
...
+      , "runCmd" : "${{config.CMD_PREFIX}} java ${{config.JVM_MEMORY}} -Dspring.aot.enabled=true -jar ${{SPRING3_BOOT_DIR}}/target/springboot3.jar"
+      },
+    "spring3-native" : {
+      "name": "spring3-native"
+      , type: "native"
+      , updateScript: "update-spring-boot-version"
+      , updateVersion: "${{config.SPRING_BOOT_VERSION}}"
+      , dir: "${{SPRING3_BOOT_DIR}}"
+      , buildCmd: "./mvnw clean -Pnative -DskipTests native:compile package ${{config.NATIVE_SPRING_BUILD_OPTIONS}}"
+      , "runCmd" : "${{config.CMD_PREFIX}} ${{SPRING3_BOOT_DIR}}/target/springboot3 ${{config.JVM_MEMORY}}"
```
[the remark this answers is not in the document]
edeandrea: Done

benchmarks/main.yml
franz1981: @edeandrea PTAL

benchmarks/helpers/os.yml
```diff
@@ -0,0 +1,76 @@
+name: OS-specific/detection scripts
```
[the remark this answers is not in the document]
edeandrea: Thats kind-of a requirement unfortunately, although you'll notice in `os.yml`:

```yaml
  sudo-linux:
    - sh: sudo ${{command}}

  sudo-macos:
    - sh: ${{command}}

  sudo:
    - script: sudo-${{os}}
      with:
        command: ${{command}}
```

on mac, sudo isn't needed. on Linux, sudo is required in order to ensure all the needed packages are installed

sudo is also needed for cgroups, but in `cgroups.yml` you'll notice that it does an os check and doesn't even try to do anything with cgroups if on a non-linux os.

so yes, if doing remote ssh, there are 2 requirements:
1. You have passwordless login set up
1. You are able to do passwordless sudo

i'm not sure of any way to get around that. i'll add that to the docs

benchmarks/helpers/requirements.yml
```diff
@@ -0,0 +1,87 @@
...
+
+  ensure-jvm-opts:
+    - read-state: config.profiler.name
+      then:
+        - regex: jfr|flamegraph
+          then:
+            - set-state: PROFILER_JVM_ARGS -XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints
+
+  ensure-gh-cli-linux:
+    - script: sudo
```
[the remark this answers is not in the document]
edeandrea: A requirement is that you have to have passwordless sudo privs set up on the host machine. Kinda sucks, but it is what it is (I don't know another way around it).

edeandrea:
> @franz1981 / @holly-cummins
> 
> i'm starting to feel like there are lots (of valid) comments on the PR. I'm also not a benchmarking expert, so there are probably specific things with regard to wrk that I may not be 100% accurate here.
> 
> Would it make more sense to merge this PR and then add specific & targeted issues to fix things up?
> 
> WDYT?

Unless either of you have any objections, I'm going to merge this.

holly-cummins:
I think the 'benchmarks' folder name could make the purpose clearer (since there are benchmark-y scripts in the top level, too. But that's something we can iterate. I guess we'll need to update the perf-lab when we do, so it's not super-trivial.

edeandrea:
> I think the 'benchmarks' folder name could make the purpose clearer (since there are benchmark-y scripts in the top level, too. But that's something we can iterate. I guess we'll need to update the perf-lab when we do, so it's not super-trivial.

Let's solve the naming thing now, as the perf lab changes haven't yet been implemented. Maybe `scripts/benchmarks` to keep everything co-located?

Or `scripts/acceptable` to go along with your naming convention in the README?

Or `scripts/perf-lab`?

edeandrea:
> Let me make sure I understand the layout. There's three modes of running, right? "Quick and dirty", "you want to do it right but only have a laptop", and "you have access to a perf lab (basically, you're John O'Hara in disguise)". 
> Q&D lives in the top-level in scripts, with instructions in the main readme. 

correct

> YWTDIR (you want to do it right) and YAJOH (you are John O'Hara) share all common scripts? But YWTDIR needs extra instructions in a readme to explain all of what's going on, because you're not John O'Hara

There isn't anything shared between Q&D and YWTDIR/YAJOH.

> and YAJOH is triggered from elsewhere, using the scripts source-controlled here?

That's correct. This repo is cloned in the perf lab and `run-benchmarks.sh` is run.

## Summary

Based on all that, maybe we should move Q&D into its own subdirectory under `scripts`?

PR #18, opened 2025-10-22
Add more parameters and wait for server availability

franz1981:
IDK @holly-cummins if that defeat the purpose of simplicity but it should be opaque to users which run the default values.

If you like the type of changes I can do the same for the other scripts (that's why is in DRAFT).

### DIFF
--- a/scripts/stress.sh
+++ b/scripts/stress.sh
@@ -1,14 +1,43 @@
 #!/bin/bash
 
-thisdir=`dirname "$0"`
+thisdir=$(dirname "$0")
+
+WRK_SETTINGS="--timeout 2s --threads 2 --connections 10 --duration 30s --rate 200"
+JVM_PARAMS="-XX:ActiveProcessorCount=8 -Xms512m -Xmx512m"
+WAIT_SERVER=${WAIT_SERVER:-true}
+
+# if no jar is provided as argument, exit and complain
+if [ "$#" -eq 0 ]; then
+  echo "Usage: $0 <path-to-jar>"
+  exit 1
+fi
 
 # make sure the port is clear before enabling halting-on-error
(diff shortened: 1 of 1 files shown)


holly-cummins:
Hmm, I feel conflicted about this. It clearly makes the scripts more powerful, but I wonder if it makes them less useful for the intended purposes:
1. A sceptical user can read the scripts, and quickly see it's just a simple wrk2/hyperfoil invocation
2. A user who has 62 other things to do can just invoke-and-go without thinking

I know the defaults mean users can still just invoke-and-go, but the extra capability does reduce the readability of the scripts. If we wanted to do things 'properly', wouldn't we use the 'medium-complexity' scripts that Eric is working on? If the crappy scripts are actually ok, it leaves less of a gap for the medium scripts to fit into. :)

franz1981:
I have mixed feelings as well.
What is the exact real purpose of the script? What is intended to demonstrate (from our pov)?

franz1981:
> here's what's wrong with the numbers you just got'. But there's no point in having two 'do it properly' sets of scripts

Thanks, got it!
In this regard I think, even before my changes, this was too good.
It set the number of cores, memory, it uses wrk2..
It even start/stop via docker a db.
Most of the time people just:
1. Run startup tests by reading what the framework log (started in 0.1 ns)
2. Run tests using work
3. Manually start stop DBs maybe not even populating them (or doing it manually, again)
4. Don't setup the number of cores
5. Uses JMeter (or wrk 1)
6. Don't configure the load generator AND the system under test
7. Perform CPU heavy work while running the test (w few cores)

edeandrea:
@franz1981 is this still needed? I dont think it does because the current `stress.sh` script has evolved since this PR was created. It waits for app startup as well as uses hyperfoil, not wrk.

PR #23, opened 2025-10-30
Restructure testing strategy to remove need for extra flag on build

holly-cummins:
We would like it so that `mvn verify` works out of the box on the Quarkus project. The reason it doesn't is because the `QuarkusIntegrationTest` runs in prod mode (as all integration tests do), and it assumes the database has been populated. We can move its tests to a `QuarkusTest`, which runs in test mode and so gets a populated database. Then, because of course we still want the `QuarkusIntegrationTest` coverage, we can make a `QuarkusIntegrationTest` which assumes the database is empty on startup. 

This setup is hard to replicate in the Spring projects, and it means the Quarkus projects have more test coverage. I think this is ok, since we want the Quarkus project to be equivalent, but also idiomatic.


edeandrea (commented):
LGTM - can you also do the same on the other quarkus project so that they are the same?

holly-cummins:
> LGTM - can you also do the same on the other quarkus project so that they are the same?

Will do. I wondered about doing it and then wasn't sure if its testing should mirror the spring-y testing or the Quarkus testing. But I guess it can't fully mirror spring so it should go all in on Quarkus.

edeandrea:
> > LGTM - can you also do the same on the other quarkus project so that they are the same?
> 
> Will do. I wondered about doing it and then wasn't sure if its testing should mirror the spring-y testing or the Quarkus testing. But I guess it can't fully mirror spring so it should go all in on Quarkus.

Until the spring testing working group completes :) Then it should mirror spring :)

holly-cummins:
> > LGTM - can you also do the same on the other quarkus project so that they are the same?
> 
> Will do. I wondered about doing it and then wasn't sure if its testing should mirror the spring-y testing or the Quarkus testing. But I guess it can't fully mirror spring so it should go all in on Quarkus.

Done! I had to make an adaptation to the application because by default the import.sql is always being loaded, even in prod mode.

I also spotted that https://github.com/quarkusio/quarkus/issues/40037 means we do not need to rename the data.sql file to import.sql on the spring project, so I renamed that file to reduce the diff to spring.

edeandrea (approved):

PR #33, opened 2025-11-05
Add instructions for manually generating spring compatibility project

holly-cummins:
This is useful for own demos, but should also be useful for people trying out the project.


holly-cummins:
I've raised https://github.com/quarkusio/spring-quarkus-perf-comparison/issues/34 (and yes please on the ownership, @edeandrea, I'd have no idea what the steps are!)

PR #71, opened 2025-11-21
Improve the stress scripts

Sanne:


holly-cummins (commented):
The changes to `stress.sh` make a lot of sense, and I think are on the right side of the 'understandability vs robustness' tradeoff. 

For `stress-all.sh`, what's the reasoning for duplicating `stress.sh` rather than just calling out to it? Is it to be able to redirect just parts of it to a file?

scripts/stress-all.sh
```diff
@@ -0,0 +1,51 @@
...
+    for ((i=0; i<30; i++)); do
+        # Using 127.0.0.1 is safer than localhost on macOS to avoid IPv6 ::1 mismatch
+        if (echo > /dev/tcp/127.0.0.1/8080) >/dev/null 2>&1; then
+            return 0
+        fi
+        sleep 1
+    done
+    echo "Timeout waiting for port 8080"
+    return 1
+}
```
Sanne: @quintesse have a look also at this script, could be useful for your own; this one waits just enough for the frameworks to have opened port 8080.  Seems like a good replacement for the "10 seconds" rule.

Also, you could then use the total time of the benchmark - including boostrap times - as a fair comparison.

edeandrea:
@holly-cummins @Sanne @franz1981 I'm trying to go through & clean up old PRs. Do we still want this one? `stress.sh` on `main` has diverged quite a bit since this PR was created.

holly-cummins:
Good question. It's been on my list all this time :( 

We (me) should tidy it up and get it merged.

PR #114, opened 2026-01-06
Configure Hikari connection pool size and set default batch fetch size

simasch:
To create the same conditions for both frameworks regarding database access, I set the fetch size and the connection pool to the same size.


franz1981:
IIRC we are still waiting the new release Quarkus version which uses 20 as default value: in theory we should make sure both default matches

PR #117, opened 2026-01-09
Practice active benchmarking

### LINKED ISSUE #62: Practice active benchmarking
A common problem while running benchmarks is to assume what's the limiting factor the benchmark is stressing and the simpler the benchmark, the easier is to believe to know the unknown.
To fix this, is highly suggested to practice https://www.brendangregg.com/activebenchmarking.html in the form of opt-in profiling/monitoring for the different components involved, whilst the tests are running 
eg
1. Enable lightweight monitoring, like sar/mpstat/time/pidstat or docker stats, against all components 
2. Opt-in to make sure that with/without such observability tools, there won't be any unexpected and unpredictable "observer" effect

As part of this, although with an higher and more biased resolution level, I would enable (still opt-in because more heavy, for the type of test) startup profiling too, which I can help setup, for both JVM and native image mode.
The latter is not trivial, but afaik, be we can ask our Mandrel team if the CPU profiler recently introduced for native-image is suitable for this job, and accurate enough, compared to Linux perf. If we enable native image profiling this is going (iirc) to impact building native too, because it requires additional compilation flags to produce debug symbols required to resolve the application frame stacks properly.

In short:
1. let's start by proving lightweight monitoring (opt-in) via sar/mpstat/time/pidstat on components 
2. Optional, but highly recommended, enable system wide, to collect similar statistics: this is very tied to our current choice to run everything local, but is good to ensure there is not overall saturation in the system due to scarse resources or noisy neighbors presence
3. Eventually, add finer grain profiling to JVM mode and native image (JVM already does it under load, but not at start-up, which is a whole different beast)

For 1, we can collect the output of monitoring and report it raw too; it doesn't need ATM to be part of the results, but still we need to make it easier to consume to detect unwanted saturations which could invalidate our results.

This is very key to make sure our results are valid and in particular, if run by others which environment is drastically different from our lab e.g. while they mine Bitcoin with their CPUs eheh

I will provide few pointers of existing qDup scrips from other OSS test we performed which could be (re)used here.

edeandrea:
- Monitor `pidstat` for the app and database during each iteration
- Monitor `mpstat` and `vmstat` for the entire machine during each iteration

All the monitoring logs are also available as part of the run.

Fixes #62

### DIFF
--- a/scripts/infra.sh
+++ b/scripts/infra.sh
@@ -13,6 +13,7 @@ help() {
   echo " -m <MEMORY>           Memory to allocate"
   echo "                         Default: ${MEMORY}"
   echo " -p <CPUSET_CPUS>      CPUs in which to allow execution (0-3, 0,1)"
+  echo " -r                    Output the PostgreSQL process host PID"
   echo " -s                    Start the services"
 }
 
@@ -20,6 +21,10 @@ exit_abnormal() {
   echo
   help
   exit 1
+}
+
+get_postgres_host_pid() {
+  echo $(${engine} inspect -f '{{.State.Pid}}' ${DB_CONTAINER_NAME})
 }
 
 # Wrapper to handle rootless podman cgroup issues on Linux
@@ -90,6 +95,7 @@ stop_postgres() {
 }
 
 start_services() {
+  echo "Using $engine to start containers"
   echo "-----------------------------------------"
   echo "[$(date +"%m/%d/%Y %T")]: Starting services"
   echo "-----------------------------------------"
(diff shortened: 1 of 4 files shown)


franz1981:
Now my friend, let's have fun eheh
Can you run a job where:
- the application has few more cores
- postgres got a single core dedicated
- hyperfoil still the same

?
And let's verify that posgres is dying and what happen to the application?
This will help to validate the effectiveness of active benchmarking for both quarkus and spring.

franz1981:
> but can we do that after this is merged?

It's ok, as we followup with fixes/improvements if the results of it won't show up the active benchmarking stuff added to be correct/enough

PR #124, opened 2026-01-09
Fix/remove open-session-in-view

### LINKED ISSUE #118: Fix open-session-in-view
Fix the open-session-in-view problem in both Spring & Quarkus versions by introducing a DTO layer and have mapping inside a transactional service layer.

edeandrea:
Fixes #118

### DIFF
--- a/quarkus3-spring-compatibility/src/main/java/org/acme/dto/AddressDTO.java
+++ b/quarkus3-spring-compatibility/src/main/java/org/acme/dto/AddressDTO.java
@@ -1,0 +1,28 @@
+package org.acme.dto;
+
+import jakarta.validation.constraints.NotBlank;
+
+public record AddressDTO(
+    @NotBlank(message = "Address is mandatory")
+    String address,
+
+    @NotBlank(message = "City is mandatory")
+    String city,
+
+    @NotBlank(message = "Country is mandatory")
+    String country
+) {
+  public AddressDTO {
+    if ((address == null) || address.isBlank()) {
+      throw new IllegalArgumentException("Address is mandatory");
+    }
+
+    if ((city == null) || city.isBlank()) {
+      throw new IllegalArgumentException("City is mandatory");
+    }
+
+    if ((country == null) || country.isBlank()) {
+      throw new IllegalArgumentException("Country is mandatory");
+    }
+  }
+}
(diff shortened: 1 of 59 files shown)


quarkus3-spring-compatibility/src/main/java/org/acme/dto/AddressDTO.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  public static class Builder {
```
franz1981: It is that common that users allocate a builder too?
Since is a CPU bound scenario I would prefer not putting even more pressure to the GC if not required

franz1981 (changes requested):

quarkus3-spring-compatibility/src/main/java/org/acme/dto/FruitDTO.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  public static class Builder {
+    private Long id;
+    private String name;
+    private String description;
+    private List<StoreFruitPriceDTO> storePrices = new ArrayList<>();
```
franz1981: Ditto as before with the problem that in theory we know upfront the size of this list and we could just allocate a List.of (which is immutable and more optimized in term of allocation size) on the record

quarkus3-spring-compatibility/src/main/java/org/acme/dto/StoreFruitPriceDTO.java
```diff
@@ -0,0 +1,53 @@
...
+    return new Builder();
+  }
+
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  public static class Builder {
+    private StoreDTO store;
+    private BigDecimal price;
```
franz1981: We could just use a primitive float/double in the record as well saving allocating full blown costly objects

edeandrea: We could, but should we? We'd potentially lose the precision (we're modeling prices in a local currency).

PR #250, opened 2026-03-03
Use SDKMAN to install Java 21

wfouche:
Install Java 21 (Temurin) with SDKMAN.

### DIFF
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -1,0 +1,7 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+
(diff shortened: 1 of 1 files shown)


wfouche:
The specified JDK will temporarily become the JDK to be used with this project. If you close the project and open another project (which doesn't have a .sdkmanrc file) then it should default again to Java 25. I find it useful, because some projects I work on require Java 17 and I just rely on .sdkmanrc and IntelliJ to select the correct JDK as I bounch between projects.

edeandrea:
i've actually just played around with it a bit. I'm fine with this change if you are @holly-cummins

PR #264, opened 2026-03-05
Reduce transaction overhead

simasch:
Hi,

It's me again. I've updated the Spring Boot 4 project to the latest version and changed the transaction propagation to reduce transaction overhead, which should improve performance.

It would be amazing if you could run the tests with my updates.

Thanks a lot.
Simon

### DIFF
--- a/quarkus3-virtual/src/main/java/org/acme/repository/FruitRepository.java
+++ b/quarkus3-virtual/src/main/java/org/acme/repository/FruitRepository.java
@@ -2,7 +2,6 @@ package org.acme.repository;
 
 import static jakarta.transaction.Transactional.TxType.SUPPORTS;
 
-import java.util.List;
 import java.util.Optional;
 
 import jakarta.enterprise.context.ApplicationScoped;
@@ -18,10 +17,4 @@ public class FruitRepository implements PanacheRepository<Fruit> {
 	public Optional<Fruit> findByName(String name) {
 		return find("name", name).firstResultOptional();
 	}
-
-  @Override
-  @Transactional(SUPPORTS)
-  public List<Fruit> listAll() {
-    return PanacheRepository.super.listAll();
-  }
 }


springboot4/pom.xml
```diff
@@ -30,10 +30,6 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-validation</artifactId>
 		</dependency>
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-actuator</artifactId>
-		</dependency>
```
edeandrea: Removing the actuator dependency makes the comparison unfair. The Quarkus version has health checks and similar capabilities. Also, most Spring Boot applications contain this, so I would consider it "unusual" to not have it.

I'm not in favor of removing this.

springboot4/src/main/java/org/acme/rest/FruitController.java
```diff
@@ -25,19 +25,19 @@ public FruitController(FruitService fruitService) {
 		this.fruitService = fruitService;
 	}
 
-	@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+	@GetMapping
```
edeandrea: Is the removal of the annotation properties causing a problem?

We have these on the Quarkus side as well.

springboot4/src/main/java/org/acme/rest/FruitController.java
```diff
@@ -25,19 +25,19 @@ public FruitController(FruitService fruitService) {
...
-	@GetMapping(path = "/{name}", produces = MediaType.APPLICATION_JSON_VALUE)
+	@GetMapping(path = "/{name}")
 	public ResponseEntity<FruitDTO> getFruit(@PathVariable String name) {
 		return this.fruitService.getFruitByName(name)
 			.map(ResponseEntity::ok)
 			.orElseGet(() -> ResponseEntity.notFound().build());
 	}
 
-	@PostMapping(produces = MediaType.APPLICATION_JSON_VALUE, consumes = MediaType.APPLICATION_JSON_VALUE)
+	@PostMapping
```
edeandrea: Is the removal of the annotation properties causing a problem?

edeandrea (commented):
Hi @simasch thanks for this! I do have some comments about some of these changes.

Notably, I'm not in favor of removing the actuator dependency because it makes the comparison unfair. Most Spring apps have this dependency, and the Quarkus application contains similar functionality that would no longer exist in the Spring version.

springboot4/src/main/java/org/acme/rest/FruitController.java
simasch: This is just redundant because JSON is the default.

springboot4/src/main/java/org/acme/rest/FruitController.java
```diff
@@ -25,19 +25,19 @@ public FruitController(FruitService fruitService) {
...
 	}
 
-	@GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+	@GetMapping
 	public List<FruitDTO> getAll() {
 		return this.fruitService.getAllFruits();
 	}
 
-	@GetMapping(path = "/{name}", produces = MediaType.APPLICATION_JSON_VALUE)
+	@GetMapping(path = "/{name}")
```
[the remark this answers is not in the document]
simasch: Having producers and/or consumers will trigger a content-type-matching mechanism in Spring, which Quarkus doesn't, I assume.

springboot4/src/main/java/org/acme/rest/FruitController.java
edeandrea: Quarkus does the same regardless of whether the produces/consumes annotations are there, and json is also the default in Quarkus.

edeandrea:
@franz1981 @holly-cummins  with this change on the spring side I'm not sure the way transactions are managed is now apples to apples. 

[Quarkus is just using the default](https://quarkus.io/guides/transaction#declarative-approach) which is `TxType.REQUIRED`

If we are going to do an apples to apples comparison based on the change made by @simasch in this PR, maybe we need to change the Quarkus app to use `@Transactional(SUPPORTS)`

My guess is that will also speed up Quarkus since it will reduce transaction overhead for Quarkus, just as it's doing for spring.

edeandrea:
I think the question is `Is it ok/a good idea to have read operations without a transaction`? Becuase this change on the spring side essentially does that.

franz1981:
Yeah agree @holly-cummins @edeandrea : as far as I understand without a transaction, you have no isolation guarantee... (phantom reads, etc etc)
Probably in Spring for this use case would be more useful to have proper transactional required + readonly ^^

edeandrea (approved):

PR #271, opened 2026-03-10
Add OpenTelemetry instrumentation across services

### LINKED ISSUE #269: Add the full observability stack to "tuned" versions
In production most users will have a full observability stack (metrics/tracing/logs/etc). Let's add that in.

edeandrea:
Integrated OpenTelemetry instrumentation into all modules, including BOM dependencies, starter configurations, and resource updates. Enhanced `FruitService` methods for span tracing, updated `infra.sh` for Otel observability stack, and adjusted `application.yml` files to enable telemetry features across services.

Fixes #269

### DIFF
--- a/springboot3/src/main/resources/application.yml
+++ b/springboot3/src/main/resources/application.yml
@@ -8,7 +8,7 @@ spring:
     hikari:
       # @simasch: Sets the jdbc pool size to be the same that Quarkus uses
       # see https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/114
-      maximum-pool-size: 20
+      maximum-pool-size: 50
   jackson:
     default-property-inclusion: non_empty
   jpa:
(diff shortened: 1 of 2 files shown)


springboot4/src/main/java/org/acme/service/FruitService.java
```diff
@@ -20,21 +23,24 @@ public FruitService(FruitRepository fruitRepository) {
     this.fruitRepository = fruitRepository;
   }
 
+  @Observed(name = "FruitService.getAllFruits")
```
brunobat: With the `opentelemetry-spring-boot-starter` you should be able to use `@WithSpan` here as well... Not sure if it's the right aproach, though.

brunobat: Sure, let's use what they recommend.

PR #284, opened 2026-03-13
Implement C-wrapper-tool for time to 1st request measurements

### LINKED ISSUE #282: Reported/Measured startup times lower than those self-reported by Quarkus
When running the test `measure-time-to-first-request` on Quarkus I observe that the `timings` reported in the `metrics.json` file are lower than the ones self-reported by Quarkus which is counter-intuitive.

## metrics for time to first request

```
      "startup": {
        "timings": [
          39,
          32,
          40,
          32,
          41,
          32,
          31,
          42,
          33,
          33
        ],
        "avStartTime": 35.5
      },
```

## Quarkus self-reported startup-time

```
❯ grep started 20260311/run-time-initialize-resource-bundles/target-host/measure-time-to-first-request-quarkus3-native-*.log | cut -d ' ' -f 14-16
started in 0.083s.
started in 0.077s.
started in 0.076s.
started in 0.076s.
started in 0.076s.
started in 0.074s.
started in 0.076s.
started in 0.077s.
started in 0.078s.
started in 0.076s.
```

One would expect the time to first request to be greater or equal to the Quarkus reported startup time, as the application would only be able to handle a request after startup.

## Thoughts

I suspect that this is a side effect of the way the time to first request is measured in

https://github.com/quarkusio/spring-quarkus-perf-comparison/blob/d70657900560d78e5875564e080c794bc4021c49/scripts/perf-lab/main.yml#L359-L361

and

https://github.com/quarkusio/spring-quarkus-perf-comparison/blob/d70657900560d78e5875564e080c794bc4021c49/scripts/perf-lab/scripts/time-to-1st-request.sh#L13-L22

The scripts first issue the command to start the Quarkus application and only later get the start-time timestamp, so there is some delay there. Furthermore, the way the timestamp is obtained (including an if-statement and an echo) further reduces the accuracy of the start and end timestamps. The curl overhead is also something we might want to take in consideration although there is probably not much we can do.

I will try to perform some profiling on my Linux workstation (were I obtained the above numbers) and update this issue with data accordingly.

zakkak:
Fixes #282

### DIFF
--- a/scripts/perf-lab/scripts/time_to_1st_request.c
+++ b/scripts/perf-lab/scripts/time_to_1st_request.c
@@ -89,6 +89,7 @@ pid_t forkme(char *args[], const char *log_path, atomic_int *ready_flag) {
 			}
 		}
 		atomic_barrier(ready_flag);
+		*((long*)ready_flag) = now_nsec();
 		execvp(args[0], args);
 		// execvp only returns if there is an error
 		exit(1);
@@ -224,9 +225,10 @@ int main(int argc, char *argv[]) {
 		return 1;
 	}
 
-	// Create shared memory for atomic synchronization flag
-	atomic_int *ready_flag = mmap(NULL, sizeof(atomic_int), PROT_READ | PROT_WRITE,
-	                               MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+	// Create shared memory for atomic synchronization flag and timestamp. We
+	// use the same memory for both, so we allocate space for the bigger one
+	atomic_int *ready_flag = mmap(NULL, sizeof(atomic_int) > sizeof(long) ? sizeof(atomic_int) : sizeof(long),
+								   PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
 	if (ready_flag == MAP_FAILED) {
 		perror("mmap");
 		free(cmd_copy);
@@ -296,6 +298,7 @@ int main(int argc, char *argv[]) {
 
 	freeaddrinfo(res);
 
+	start_time = *((long*)ready_flag);
 	printf("http_code=%d attempts=%d elapsed=%ld ns\n", code, attempts, end_time - start_time);
 
 	// Clean up: kill child process


scripts/perf-lab/main.yml
```diff
@@ -355,22 +355,15 @@ scripts:
       then:
         - queue-download: ${{REPO_DIR}}/logs/measure-time-to-first-request-${{RUNTIME.name}}-${{ITERATION}}.log
         - script: start-test-services
+        - sh: make -C ${{HELPER_SCRIPTS_DIR}}
         - script: sync-drop-fs-cache
-        - sh: ${{RUNTIME.runCmd}} &>${{REPO_DIR}}/logs/measure-time-to-first-request-${{RUNTIME.name}}-${{ITERATION}}.log &
-        - sh: APP_PID=$!
-        - sh: timeout 60s bash -c "taskset --cpu-list ${{config.resources.cpu.1st_request}} ${{HELPER_SCRIPTS_DIR}}/time-to-1st-request.sh ${{TARGET_URL}}"
+        - sh: ${{APP_CMD_PREFIX}} ${{HELPER_SCRIPTS_DIR}}/time_to_1st_request "${{RUNTIME.runCmd}}" "${{REPO_DIR}}/logs/measure-time-to-first-request-${{RUNTIME.name}}-${{ITERATION}}.log" "${{TARGET_URL}}" 60
```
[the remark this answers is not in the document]
franz1981: And with quarkus we cannot afford to taskset it later, because tons of JVM ergonomics depends by the number if core perceives at startup

scripts/perf-lab/scripts/time_to_1st_request.c
```diff
@@ -0,0 +1,310 @@
...
+			// Redirect stdout and stderr to log file
+			int log = open(log_path, O_WRONLY | O_CREAT | O_TRUNC, 0644);
+			if (log != -1) {
+				dup2(log, STDOUT_FILENO);
+				dup2(log, STDERR_FILENO);
+				close(log);
+			}
+		}
+		atomic_barrier(ready_flag);
+		execvp(args[0], args);
```
franz1981: The barrier ensures that both the child and the parent "go" at the same time, but start_time is recorded by the parent **after** the barrier. This means the measurement doesn't  start at the exact moment before execvp() — it starts slightly later, after the parent's own post-barrier overhead (now_nsec() + entering the loop + socket()).                                    

On very fast startups, this overhead could matter.

A more precise approach: have the child write now_nsec() into shared memory right before calling execvp(), and have the parent use that value as the start time. 
This way the measurement is anchored to the child's perspective, and the parent's post-barrier overhead becomes irrelevant.

edeandrea: All of this is making me appreciate Java's higher level concurrency constructs :)

zakkak:
`bash` implementation proposed in https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/317

PR #330, opened 2026-03-24
Add `count_cpus` function to calculate CPU count from taskset --cpu-list spec in benchmarking script

### LINKED ISSUE #328: Compute `app_cpus` field
The meaning of the `app_cpus` field has changed. We need to put in some logic here so that it gets computed as part of the scripts to remain the absolute number of CPUs that the application used in its run.

See
- https://github.com/quarkusio/benchmarks/pull/355
- https://github.com/quarkusio/benchmarks/pull/354

edeandrea:
Fixes #328


holly-cummins:
Looks good! 

Is it possible to add some tests? Maybe it's not needed, if we've tested it locally and are unlikely to change it. I'm thinking something like another bash script that calls that function with a few examples we want to make sure work, and then some extra steps in the actions yaml to call the little mini-harness. We might find it useful for some other things, too, as the logic gets more complex.

PR #333, opened 2026-03-25
Fix logic for reading HTTP response status code in time-to-1st-request.sh

edeandrea:
`time-to-1st-request.sh` was not the same on the ootb branch

### DIFF
--- a/scripts/perf-lab/scripts/time-to-1st-request.sh
+++ b/scripts/perf-lab/scripts/time-to-1st-request.sh
@@ -75,7 +75,7 @@ function _date() {
         continue
       fi
       # Read the HTTP response status line and extract the status code
-      if read -r _ status_code _ <&3; then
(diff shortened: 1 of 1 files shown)


edeandrea:
I've been thinking about that for a bit. I'm not sure we'd want to do that for the apps themselves because they are different now with different pieces, but the scripts & everything that executes the benchmarks are (currently) identical.

How long to they stay that way? Not sure, but I supposed at some point in time it will happen. So maybe separating the scripts from the apps is a good idea? Although that adds cognitive overhead, because now the scripts are somewhere different.

edeandrea:
OOh ooh ooh I just thought of something? What if we had a bot that didn't do anything automatically, but we could tag it.

Something like `@automerge-bot merge this PR to ootb branch`

And we could tag the PR based on `app`, `infra`, `scripts`, etc. And the merge bot would only merge the things based on tags.

PR #422, opened 2026-04-09
Fix syncbot workflow to run on pull_request events

diegolovison:
Fix https://github.com/quarkusio/spring-quarkus-perf-comparison/issues/415

### DIFF
--- a/.github/workflows/syncbot.yml
+++ b/.github/workflows/syncbot.yml
@@ -9,7 +9,6 @@ on:
 jobs:
   prep-changes:
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
     steps:
       - uses: actions/checkout@v6
 


edeandrea:
> I've been thinking about how we only run trusted workflows. I can think of a few options:
> 
> * Put a wait and then an extra check in the syncbot to not raise a PR until it can see a workflow is running for the original PR (it can't use the workflow as the trigger because it won't run, to avoid workflow storms)
> * Adjust our repo settings to always require manual workflow approval for the syncbot user, but that would add an extra, annoying, step
> * Check the permissions of the author of the original PR (although I'm not sure how easy that is), and then ... maybe mark the child PR as draft if it's not from a trusted author? Ideally we'd make the PR itself require manual approval but I don't think github allows that

I know we wanted more of an "opt out" vs "opt in" approach, but maybe we go back to the labelling thing? Someone with correct permissions to assign labels (triage?) has to label the PR to be "synced", and that is what gives syncbot the permission to do so? That way someone with the permissions has reviewed and said "this PR is ok"?

holly-cummins (approved):

holly-cummins:
The CI for this is now in a bit of a mess, because I've rebased several times and it's having conflicts while cherry-picking. That situation *is* something that should be handled, but I don't think it needs to be handled before merging this PR. I think it's an issue that will mostly affect PRs that stay open for a while and get rebased several times.

PR #424, opened 2026-04-09
Do not assume target branch is on the same fork as the CI

holly-cummins:
Should fix 

```
fatal: ambiguous argument 'origin/ootb..origin/fix_415': unknown revision or path not in the working tree.
```

happening in the PR workflow for #422.


edeandrea (approved):

edeandrea:
In order for an automation to run on a PR branch it has to exist on the branch that the PR is targeting. That's why there is a separate `.github/workflows/main.yml` on the `ootb` branch.

PR #433, opened 2026-04-09
Log firewall status

diegolovison:


scripts/perf-lab/helpers/os.yml
```diff
@@ -34,6 +34,9 @@ scripts:
       then:
         - download: ${{REPO_DIR}}/logs/lscpu.txt
         - sh: cat ${{REPO_DIR}}/logs/lscpu.txt
+    - sh: sudo systemctl status firewalld > ${{REPO_DIR}}/logs/nerd-stats.txt 2>&1 &
```
edeandrea: There's a `sudo` script - shouldn't we call that instead?

scripts/perf-lab/helpers/os.yml
```diff
@@ -34,6 +34,9 @@ scripts:
...
+    - sh: tuned-adm active >> ${{REPO_DIR}}/logs/nerd-stats.txt 2>&1 &
```
edeandrea: Seems we're re-using `nerd-stats.txt` for multiple things. Should we call it what it really is and dump each thing out to its own file?

diegolovison: Well, I was thinking about this. Maybe we should append to metrics.json?

diegolovison: I would like to have a single file. Downloading a lot of files is easy to get lost.

edeandrea: > Well, I was thinking about this. Maybe we should append to metrics.json?

I don't think so. `metrics.json` is for the metrics, not system logging. We output `lscpu` to its own file, so I'm not sure why we wouldn't follow the same pattern here?

edeandrea: > I would like to have a single file. Downloading a lot of files is easy to get lost.

These files are also published publicly, so if multiple things are in a single file called `nerd-stats`, how do people actually know what info is in those files.

I think keeping separate files for each piece of info is prefereable.

WDYT @holly-cummins ?

holly-cummins: Argh, unhelpfully, I don't know! I like how if there's lots of files you can scan the names to see what you'll find, but I like how in a single file, the timestamp-correlation is already done for you, and you only need to download one thing. Also, I do kind of like the `nerd-stats` name :)

scripts/perf-lab/helpers/os.yml
holly-cummins: I decided this would be the easiest change to address to test sync update, but I can't find example usages of the sudo script and my qdup isn't good enough to switch without an example to follow.

scripts/perf-lab/helpers/os.yml
edeandrea: > Also, I do kind of like the nerd-stats name :)

I do too :) But if I'm unfamiliar with this repo and I stumble upon it from a blog post or something, I would think that having all the log files in a single folder named after what's in them would be most helpful?

The files are already correlated together for the run (https://github.com/quarkusio/benchmarks/tree/main/results/spring-quarkus-perf-comparison/2026-03-25_16-16-36 for instance).

scripts/perf-lab/helpers/os.yml
edeandrea:
```yml
- script: sudo
  with:
    command: systemctl status firewalld > ${{REPO_DIR}}/logs/nerd-stats.txt 2>&1 &
```

edeandrea: https://github.com/quarkusio/spring-quarkus-perf-comparison/blob/3fc38ea283e7e56a68218c1deac1d09b099d9c5b/scripts/perf-lab/helpers/os.yml#L90-L99 is an example

holly-cummins:
```suggestion
    - script: sudo
       with: systemctl status firewalld > ${{REPO_DIR}}/logs/nerd-stats.txt 2>&1 &
```

scripts/perf-lab/helpers/os.yml
diegolovison: Let's keep separated. Then we have a pattern.

scripts/perf-lab/helpers/os.yml
```diff
@@ -34,6 +34,8 @@ scripts:
       then:
         - download: ${{REPO_DIR}}/logs/lscpu.txt
         - sh: cat ${{REPO_DIR}}/logs/lscpu.txt
+    - sh: sudo systemctl status firewalld > ${{REPO_DIR}}/logs/firewall-status.txt 2>&1
```
edeandrea: Shouldn't this use the `sudo` script rather than call `sudo` directly?

```yaml
- script: sudo
  with:
    command: systemctl status firewalld > ${{REPO_DIR}}/logs/firewalld-status.txt 2>&1
```

scripts/perf-lab/helpers/os.yml
```diff
@@ -34,6 +34,8 @@ scripts:
...
+    - download: ${{REPO_DIR}}/logs/firewall-status.txt
```
edeandrea: `firewalld-status.txt`?

scripts/perf-lab/helpers/os.yml
diegolovison: Done

scripts/perf-lab/helpers/os.yml
diegolovison: Done

diegolovison:
@edeandrea Done. Already tested on build #299

scripts/infra.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -euxo pipefail
```
edeandrea, line 2: Why the addition of `x`?

diegolovison: When running the `infra.sh` script, without `x`, you cannot see what is being executed. If you need to reproduce something or troubleshoot, you never know what the script executed.

edeandrea: Right. I understand that, but this script gets executed a lot throughout a run. Wouldn't that lead to a lot of noise?

diegolovison: For me, no. It is a log file with just a few more lines. But if you prefer not having, I don't have a strong opinion here.

PR #436, opened 2026-04-10
Handle conflicting cherry-picks in syncbot, also reduce conflicts by not making every PR try to fully align the branches

holly-cummins:
I'm seeing conflicts in the syncbot even for PRs which haven't been open for very long. This surprises me, but I think the right thing to do is skip-and-move-on, not die-horribly.

- Skip conflicting cherry-picks instead of aborting the entire sync
- Still create the sync PR with successfully cherry-picked commits
- Note skipped commits in the sync PR body and comment on the original PR

*Edit*: Ohh, the reason there are conflicts is that it's trying to cherry-pick every change, not just the ones in this PR. That's bad. I've updated this change to also fix that.

### DIFF
--- a/.github/workflows/syncbot.yml
+++ b/.github/workflows/syncbot.yml
@@ -60,6 +60,7 @@ jobs:
           
           # Cherry-pick each commit
           echo "Cherry-picking commits..."
+          SKIPPED_COMMITS=""
           for commit in $COMMITS; do
             echo "Cherry-picking $commit"
             git cherry-pick $commit || {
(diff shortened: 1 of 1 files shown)


edeandrea:
What does it mean "Some commits were skipped due to conflicts and may need manual syncing. 🔄"? Does that really even make sense? If a PR synced some commits but not others, couldn't that just wind up in a big mess?

I wonder if just notifying the user that they need to manually sync the PR be better?

holly-cummins:
> What does it mean "Some commits were skipped due to conflicts and may need manual syncing. 🔄"? Does that really even make sense? If a PR synced some commits but not others, couldn't that just wind up in a big mess?

My expectation is that a person would go in and update the PR, or make their own, but that it's no worse a mess than if they'd try to do the cherry-pick themselves. 

> I wonder if just notifying the user that they need to manually sync the PR be better?

I wondered about that, but I think actually having the PR is kind of useful, because it has a diff that the user can look at. If the bot just put a comment on the original PR, it might be easier to miss, because there's no PR demandsing attention. You wouldn't be able to see the commits that did merge ok. 

Although once I fixed the "the point of the PR isn't to make the two branches identical" issue, we should hardly ever get merge conflicts. When we do, they're probably because a change shouldn't apply. I'm also not sure what this would do in the one-commit case ... it might still fail.

edeandrea (approved):

holly-cummins:
We can iterate if we find the conflict behaviour is annoying... once it's working (eventually? :) )

PR #440, opened 2026-04-10
Improve job name in syncbot CI

holly-cummins:
Raising this partly because I don't like the name of the job, but mostly because I think trying to mirror a change in this file will be carnage, and I want to see in what way the syncbot dies.

### DIFF
--- a/.github/workflows/syncbot.yml
+++ b/.github/workflows/syncbot.yml
@@ -11,7 +11,7 @@ permissions:
   pull-requests: write
 
 jobs:
-  prep-changes:
+  mirror-changes:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6


holly-cummins:
```
 Successfully pushed sync branch
pull request create failed: GraphQL: No commits between ootb and sync-improve-job-naming-to-ootb (createPullRequest)
Error: Process completed with exit code 1.
```

Not quite what I expected, but fine, I guess?

edeandrea:
What does that actually mean?

holly-cummins:
I think it's saying that it couldn't see any difference. Whereas I'd expect it to say it couldn't find the file. I guess we keep an eye on it.

holly-cummins:
Just now I tried to do a manual PR to get the syncbot file into the ootb branch and instead just pushed it into the branch (ooops), so if we tried again on this specific PR I'd definitely expect it to say there were no changes.

PR #473, opened 2026-04-17
Add energy measurement support (RAPL + iDRAC)

### LINKED ISSUE #99: Measure energy in perf lab scripts
Now that we have the automation working (yay!), we can go to the next step and start measuring energy as well. 

I'm assuming we'll replicate what John O'Hara did and access RAPL directly. I don't know if we have access to his actual scripts for that part anymore. Alternatively, for more portability, we could use the approach from #38. That has the advantage of not being restricted to intel hardware, but the disadvantage that it's introducing another moving part into the process, and thus another bit of methodology that needs validation.

holly-cummins:
Resolves #99.

This is based on John O'Hara's work in the `quarkus-qe-spring-boot-comparison` branch of his old repo. I've copied the relevant qDup scripts from his yaml, and then added some extra logic to allow it to be turned on and off, guard execution on non-Intel hardware and non-Dell hardware, and report an average energy consumption into the json. 

The energy measurements will run as part of the same job that gathers rss information. That seems suitable, since it's fixed throughput, and we're running it already. (@franz1981 and @edeandrea, should that job be made longer, to get higher quality results?)

I'm stumbling in the dark a bit with the qDup, since this is my first time touching the scripts. I've tested as well as I can, but I don't have access to any Intel hardware, so I had to test by getting claude to stub out the rapl and idrac endpoints. That's not brilliant, so I'd be ever so grateful if @edeandrea or @franz1981 could test for me. :) 

These are the changes:

- Add RAPL (CPU-level) and iDRAC (system-level) power measurement to the `measure-rss` benchmark phase
 - New --energy-rapl and --energy-idrac CLI flags (enabled|disabled, default: enabled) to control which measurements run                           
  - Hardware guards silently skip unsupported measurements (e.g. no RAPL on ARM, no racadm on non-Dell), so both can safely be left enabled on any  platform  
- New `scripts/perf-lab/helpers/energy.yml` with `install-rapl-plot` and `capture-system-power` scripts

### Energy measurement outputs
- `.rapl` files — per-second CPU package and DRAM power readings from Intel RAPL
- `.svrPwr` files — per-second system-level wattage readings from Dell iDRAC/racadm
- new stanzas in the json. Claude claims they will look like this, with per-iteration average watts for each source, plus cross-iteration averages.                                    

```                                                                                                                                                    
  "energy": {                                                                                                                                       
    "idrac": [157, 155],                                                                                                                            
    "avIdrac": 156,                                                                                                                                 
    "raplPkg": [25.85, 26.10],                                                                                                                      
    "avRaplPkg": 25.975,                                                                                                                            
    "raplDram": [5.42, 5.38],                                                                                                                     
    "avRaplDram": 5.40                                                                                                                              
  }                                                                                                                                               
                                                                                                                                                   
```

### DIFF
--- a/scripts/perf-lab/helpers/energy.yml
+++ b/scripts/perf-lab/helpers/energy.yml
@@ -17,6 +17,35 @@ scripts:
                   with:
                     array: RUN.svrPwr
                     value: ${{watts}}
+  idrac-login:
+    - sh: IDRAC_HOST="${{idrac}}"
+    - sh: IDRAC_USER="${{idrac_user:root}}"
+    - sh: IDRAC_PWD="${{password}}"
+    - sh: REAL_RACADM=$(which racadm)
+    - sh: 'racadm() { "$REAL_RACADM" -r "$IDRAC_HOST" -u "$IDRAC_USER" -p "$IDRAC_PWD" --nocertwarn "$@"; }'
+    - sh: racadm getsysinfo
+      then:
+        - regex: "System Model"
+          then:
+            - log: Connected to iDRAC at ${{idrac}}
+          else:
+            - abort: "Failed to connect to iDRAC at ${{idrac}}. Verify hostname and credentials."
+
+  validate-idrac:
+    - read-state: ${{config.energy.idrac}}
+      then:
+        - regex: "enabled"
+          then:
+            - sh: "which racadm &>/dev/null && echo RACADM_FOUND || echo RACADM_MISSING"
+              then:
+                - regex: RACADM_FOUND
+                  then:
+                    - set-state: RUN.IDRAC_AVAILABLE true
+                - regex: RACADM_MISSING
+                  then:
+                    - set-state: RUN.IDRAC_AVAILABLE false
+                    - log: racadm not found on this system, skipping energy measurement via iDRAC
+
   install-rapl-plot:
     - read-state: ${{config.energy.rapl}}
       then:
--- a/scripts/perf-lab/helpers/energy.yml
+++ b/scripts/perf-lab/helpers/energy.yml
@@ -1,0 +1,42 @@
+name: Energy-related scripts
+scripts:
+  capture-system-power:
+    - script: idrac-login
+      with:
+        idrac: ${{= ${{servers}}.filter( server => { return server.hostname == "${{env.HOST}}"})[0].idrac}}
+        password: ${{idrac_pwd}}
+    - set-state: RUN.svrPwr []
+    - repeat-until: ${{COMPLETE_SIGNAL}}
+      then:
+        - sleep: 1s
+        - sh: racadm get system.Power.Realtime.Power
+          then:
+            - regex: (?<watts>\\d+) W | (?<btu>\\d+) Btu/hr
+              then:
+                - script: state-array-push
+                  with:
+                    array: RUN.svrPwr
+                    value: ${{watts}}
+  install-rapl-plot:
+    - read-state: ${{config.energy.rapl}}
+      then:
+        - regex: "enabled"
+          then:
+            - sh: "[[ -d /sys/class/powercap/intel-rapl ]] && echo RAPL_SUPPORTED || echo RAPL_UNSUPPORTED"
+              then:
+                - regex: RAPL_SUPPORTED
+                  then:
+                    - sh: "[[ -f ${{RAPL_PATH:/tmp/rapl}}/${{RAPL_EXE:rapl-plot}} ]] || echo GETRAPL;"
+                      then:
+                        - regex: GET
+                          then:
+                            - sh: mkdir -p ${{RAPL_PATH:/tmp/rapl}}
+                            - sh: cd ${{RAPL_PATH:/tmp/rapl}}
+                            - sh: git clone https://github.com/deater/uarch-configure.git
+                            - sh: cd uarch-configure/rapl-read
+                            - sh: make all
+                            - sh: cp rapl-plot ${{RAPL_PATH:/tmp/rapl}}/${{RAPL_EXE:rapl-plot}}
+                    - set-state: RUN.RAPL_BIN ${{RAPL_PATH:/tmp/rapl}}/${{RAPL_EXE:rapl-plot}}
+                - regex: RAPL_UNSUPPORTED
+                  then:
+                    - log: RAPL not available on this system, skipping energy measurement via RAPL
--- a/scripts/perf-lab/helpers/energy.yml
+++ b/scripts/perf-lab/helpers/energy.yml
@@ -107,6 +107,9 @@ scripts:
             - set-state:
                 key: ${{result_prefix}}.totalIdrac
                 value: ${{= ${{${{result_prefix}}.idrac}}.reduce((a, b) => a + b, 0) || 0 }}
+            - set-state:
+                key: ${{result_prefix}}.energyPerRequestIdrac
+                value: ${{= (${{${{result_prefix}}.idrac}}.reduce((a, b) => a + b, 0) || 0) / (${{av_throughput}} * ${{duration}} * ${{num_iterations}}) }}
     - read-state: ${{config.energy.rapl}}
       then:
         - regex: "enabled"
@@ -121,6 +124,9 @@ scripts:
                     - set-state:
                         key: ${{result_prefix}}.totalRaplDram
                         value: ${{= ${{${{result_prefix}}.raplDram}}.reduce((a, b) => a + b, 0) || 0 }}
+                    - set-state:
+                        key: ${{result_prefix}}.energyPerRequestRaplPkg
+                        value: ${{= (${{${{result_prefix}}.raplPkg}}.reduce((a, b) => a + b, 0) || 0) / (${{av_throughput}} * ${{duration}} * ${{num_iterations}}) }}
 
   install-rapl-plot:
     - read-state: ${{config.energy.rapl}}
--- a/scripts/perf-lab/helpers/energy.yml
+++ b/scripts/perf-lab/helpers/energy.yml
@@ -38,6 +38,90 @@ scripts:
                     - set-state: RUN.IDRAC_AVAILABLE false
                     - log: racadm not found on this system, skipping energy measurement via iDRAC
 
+  start-energy-capture:
+    - read-state: ${{config.energy.rapl}}
+      then:
+        - regex: "enabled"
+          then:
+            - sh: "[[ -x ${{RUN.RAPL_BIN:__NORAPL__}} ]] && echo RAPL_READY"
+              then:
+                - regex: RAPL_READY
+                  then:
+                    - script: sudo
+                      with:
+                        command: ${{RUN.RAPL_BIN}} -p > ${{RUN.OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl &
+                    - sh: echo $!
+                      then:
+                        - set-state: RUN.RAPL_PID
+    - read-state: ${{RUN.IDRAC_AVAILABLE}}
+      then:
+        - regex: "true"
+          then:
+            - set-signal: ${{COMPLETE_SIGNAL}} 1
+            - script:
+                name: capture-system-power
+                async: true
+
+  stop-energy-capture:
+    - read-state: ${{RUN.IDRAC_AVAILABLE}}
+      then:
+        - regex: "true"
+          then:
+            - signal: ${{COMPLETE_SIGNAL}}
+            - sh: echo "${{RUN.svrPwr}}" > ${{RUN.OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.svrPwr
+            - set-state: IDRAC_ENERGY ${{= ${{RUN.svrPwr}}.reduce((a, b) => a + b, 0) || 0 }}
+            - script: state-array-push
+              with:
+                array: ${{result_prefix}}.idrac
+                value: ${{IDRAC_ENERGY}}
+    - read-state: ${{config.energy.rapl}}
+      then:
+        - regex: "enabled"
+          then:
+            - sh: "[[ -x ${{RUN.RAPL_BIN:__NORAPL__}} ]] && echo RAPL_READY"
+              then:
+                - regex: RAPL_READY
+                  then:
+                    - script: sudo
+                      with:
+                        command: kill ${{RUN.RAPL_PID}}
+                    - sh: sleep 1
+                    - sh: awk -F', ' '/^[^#]/{s+=$2; d+=$4} END{printf "%.2f %.2f", s, d}' ${{RUN.OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl
+                      then:
+                        - regex: (?<pkg>\S+) (?<dram>\S+)
+                          then:
+                            - script: state-array-push
+                              with:
+                                array: ${{result_prefix}}.raplPkg
+                                value: ${{pkg}}
+                            - script: state-array-push
+                              with:
+                                array: ${{result_prefix}}.raplDram
+                                value: ${{dram}}
+
+  calc-energy-totals:
+    - read-state: ${{RUN.IDRAC_AVAILABLE}}
+      then:
+        - regex: "true"
+          then:
+            - set-state:
+                key: ${{result_prefix}}.totalIdrac
+                value: ${{= ${{${{result_prefix}}.idrac}}.reduce((a, b) => a + b, 0) || 0 }}
+    - read-state: ${{config.energy.rapl}}
+      then:
+        - regex: "enabled"
+          then:
+            - sh: "[[ -x ${{RUN.RAPL_BIN:__NORAPL__}} ]] && echo RAPL_READY"
+              then:
+                - regex: RAPL_READY
+                  then:
+                    - set-state:
+                        key: ${{result_prefix}}.totalRaplPkg
+                        value: ${{= ${{${{result_prefix}}.raplPkg}}.reduce((a, b) => a + b, 0) || 0 }}
+                    - set-state:
+                        key: ${{result_prefix}}.totalRaplDram
+                        value: ${{= ${{${{result_prefix}}.raplDram}}.reduce((a, b) => a + b, 0) || 0 }}
+
   install-rapl-plot:
     - read-state: ${{config.energy.rapl}}
       then:
--- a/scripts/perf-lab/helpers/energy.yml
+++ b/scripts/perf-lab/helpers/energy.yml
@@ -2,9 +2,6 @@ name: Energy-related scripts
 scripts:
   capture-system-power:
     - script: idrac-login
-      with:
-        idrac: ${{= ${{servers}}.filter( server => { return server.hostname == "${{env.HOST}}"})[0].idrac}}
-        password: ${{idrac_pwd}}
     - set-state: RUN.svrPwr []
     - repeat-until: ${{COMPLETE_SIGNAL}}
       then:
@@ -18,18 +15,13 @@ scripts:
                     array: RUN.svrPwr
                     value: ${{watts}}
   idrac-login:
-    - sh: IDRAC_HOST="${{idrac}}"
-    - sh: IDRAC_USER="${{idrac_user:root}}"
-    - sh: IDRAC_PWD="${{password}}"
-    - sh: REAL_RACADM=$(which racadm)
-    - sh: 'racadm() { "$REAL_RACADM" -r "$IDRAC_HOST" -u "$IDRAC_USER" -p "$IDRAC_PWD" --nocertwarn "$@"; }'
     - sh: racadm getsysinfo
       then:
         - regex: "System Model"
           then:
-            - log: Connected to iDRAC at ${{idrac}}
+            - log: Connected to local iDRAC
           else:
-            - abort: "Failed to connect to iDRAC at ${{idrac}}. Verify hostname and credentials."
+            - abort: "Failed to connect to local iDRAC. Is this a Dell server with racadm installed?"
 
   validate-idrac:
     - read-state: ${{config.energy.idrac}}
--- a/scripts/perf-lab/main.yml
+++ b/scripts/perf-lab/main.yml
@@ -17,6 +17,8 @@ states:
   config.units.energy.idrac: J
   config.units.energy.raplPkg: J
   config.units.energy.raplDram: J
+  config.units.energy.energyPerRequestIdrac: J/req
+  config.units.energy.energyPerRequestRaplPkg: J/req
   config.units.energy.throughput: tps
 
   config.jvm.home: #/path/to/java
@@ -582,8 +584,8 @@ scripts:
         - queue-download: ${{REPO_DIR}}/logs/wrk-${{RUNTIMECMD.name}}-${{ITERATION}}.log
         - set-state: RUN.OUTPUT_DIR ${{REPO_DIR}}/logs
         - set-state: RUN.COMPLETE_SIGNAL load-test-energy-${{RUNTIMECMD.name}}-${{ITERATION}}
-        - queue-download: ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl
-        - queue-download: ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.svrPwr
+        - queue-download: ${{RUN.OUTPUT_DIR}}/${{RUN.COMPLETE_SIGNAL}}.rapl
+        - queue-download: ${{RUN.OUTPUT_DIR}}/${{RUN.COMPLETE_SIGNAL}}.svrPwr
         - script: start-test-services
         - script: sync-drop-fs-cache
         - set-signal: LOAD_STEADY_STATE_START 1
@@ -726,9 +728,13 @@ scripts:
     - script: calc-energy-totals
       with:
         result_prefix: RUN.output.results.${{RUNTIMECMD.name}}.energyMaxLoad
+        av_throughput: ${{RUN.output.results.${{RUNTIMECMD.name}}.load.avThroughput}}
+        duration: 30
+        num_iterations: ${{config.num_iterations}}
 
   measure-energy-fixed-throughput:
     - log: Measuring energy at fixed throughput (${{config.energy.fixed.rate}} tps)
+    - sh: java -version
     - sh: cd ${{RUNTIMECMD.dir}}
     - set-state:
         key: APP_LOG_REGEX
@@ -744,8 +750,8 @@ scripts:
         - queue-download: ${{LOG_FILE}}
         - queue-download: ${{REPO_DIR}}/logs/wrk-energy-fixed-warmup-${{RUNTIMECMD.name}}-${{ITERATION}}.log
         - queue-download: ${{REPO_DIR}}/logs/wrk-energy-fixed-${{RUNTIMECMD.name}}-${{ITERATION}}.log
-        - queue-download: ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl
-        - queue-download: ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.svrPwr
+        - queue-download: ${{RUN.OUTPUT_DIR}}/${{RUN.COMPLETE_SIGNAL}}.rapl
+        - queue-download: ${{RUN.OUTPUT_DIR}}/${{RUN.COMPLETE_SIGNAL}}.svrPwr
         - script: start-test-services
         - script: sync-drop-fs-cache
         - set-signal: LOG_REGEX_REACHED 1
@@ -753,6 +759,12 @@ scripts:
         - sh: PID=$!
         - sh: echo $PID
         - set-state: JVM_PID
+        - script: monitor-processes
+          with:
+            pid: ${{JVM_PID}}
+            log_dir: ${{REPO_DIR}}/logs
+            runtime_name: ${{RUNTIMECMD.name}}
+            iteration: ${{ITERATION}}
         - script:
             name: watch-log
             async: true
@@ -763,7 +775,7 @@ scripts:
         - sh: ${{RUN.WRK_BIN}} -t 2 -c 100 -d 30s -R ${{config.energy.fixed.rate}} --timeout 30s ${{TARGET_URL}} 2>&1 >${{REPO_DIR}}/logs/wrk-energy-fixed-warmup-${{RUNTIMECMD.name}}-${{ITERATION}}.log
         - sleep: 10s
         - script: start-energy-capture
-        - sh: ${{RUN.WRK_BIN}} -t 2 -c 100 -d ${{config.energy.fixed.duration}}s -R ${{config.energy.fixed.rate}} --timeout ${{config.energy.fixed.duration}}s ${{TARGET_URL}} > ${{REPO_DIR}}/logs/wrk-energy-fixed-${{RUNTIMECMD.name}}-${{ITERATION}}.log 2>&1 &
+        - sh: ${{RUN.WRK_BIN}} -t 2 -c 100 -d ${{config.energy.fixed.duration}}s -R ${{config.energy.fixed.rate}} --timeout 30s ${{TARGET_URL}} > ${{REPO_DIR}}/logs/wrk-energy-fixed-${{RUNTIMECMD.name}}-${{ITERATION}}.log 2>&1 &
         - sh: WRK_PID=$! && echo $WRK_PID
         - set-state: WRK_PID
         - sh: wait ${{WRK_PID}}
@@ -787,6 +799,9 @@ scripts:
     - script: calc-energy-totals
       with:
         result_prefix: RUN.output.results.${{RUNTIMECMD.name}}.energyFixedLoad
+        av_throughput: ${{RUN.output.results.${{RUNTIMECMD.name}}.energyFixedLoad.avThroughput}}
+        duration: ${{config.energy.fixed.duration}}
+        num_iterations: ${{config.num_iterations}}
 
   watch-log:
     - sh: tail -f ${{log_file}}
--- a/scripts/perf-lab/main.yml
+++ b/scripts/perf-lab/main.yml
@@ -14,9 +14,9 @@ states:
   config.units.load.throughputDensity: tps per MiB
   config.units.load.errors.connectionErrors: absolute number
   config.units.load.errors.requestTimeouts: absolute number
-  config.units.energy.idrac: W
-  config.units.energy.raplPkg: W
-  config.units.energy.raplDram: W
+  config.units.energy.idrac: J
+  config.units.energy.raplPkg: J
+  config.units.energy.raplDram: J
 
   config.jvm.home: #/path/to/java
   config.jvm.version: 25.0.2-tem
@@ -47,6 +47,7 @@ states:
 
   config.energy.rapl: enabled #disabled
   config.energy.idrac: enabled #disabled
+  config.energy.rss.duration: 150
   config.profiler.name: none #jfr syncjfr flamegraph
   config.profiler.events: cpu
 
@@ -575,17 +576,18 @@ scripts:
           with:
             array: RUN.output.results.${{RUNTIMECMD.name}}.rss.firstRequest
             value: ${{= ${{FIRST_REQUEST}}/1024 }}
+        - sh: sleep ${{config.energy.rss.duration}}
         - read-state: ${{RUN.IDRAC_AVAILABLE}}
           then:
             - regex: "true"
               then:
                 - signal: ${{COMPLETE_SIGNAL}}
                 - sh: echo "${{RUN.svrPwr}}" > ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.svrPwr
-                - set-state: IDRAC_AVG_WATTS ${{= (${{RUN.svrPwr}}.reduce((a, b) => a + b, 0) / ${{RUN.svrPwr}}.length) || 0 }}
+                - set-state: IDRAC_ENERGY ${{= ${{RUN.svrPwr}}.reduce((a, b) => a + b, 0) || 0 }}
                 - script: state-array-push
                   with:
                     array: RUN.output.results.${{RUNTIMECMD.name}}.energy.idrac
-                    value: ${{IDRAC_AVG_WATTS}}
+                    value: ${{IDRAC_ENERGY}}
         - read-state: ${{config.energy.rapl}}
           then:
             - regex: "enabled"
@@ -598,7 +600,7 @@ scripts:
                           with:
                             command: kill ${{RUN.RAPL_PID}}
                         - sh: sleep 1
-                        - sh: awk -F', ' '/^[^#]/{s+=$2; d+=$4; n++} END{if(n>0) printf "%.2f %.2f", s/n, d/n; else print "0 0"}' ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl
+                        - sh: awk -F', ' '/^[^#]/{s+=$2; d+=$4} END{printf "%.2f %.2f", s, d}' ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl
                           then:
                             - regex: (?<pkg>\S+) (?<dram>\S+)
                               then:
--- a/scripts/perf-lab/main.yml
+++ b/scripts/perf-lab/main.yml
@@ -525,20 +525,15 @@ scripts:
         var-name: RUN.output.results.${{RUNTIMECMD.name}}.startup.avStartTime
         array: RUN.output.results.${{RUNTIMECMD.name}}.startup.timings
 
-  measure-rss: # also measures energy
+  measure-rss:
     - log: Measuring RSS
     - sh: java -version
     - sh: cd ${{RUNTIMECMD.dir}}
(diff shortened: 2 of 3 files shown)


scripts/perf-lab/run-benchmarks.sh
```diff
@@ -274,6 +274,7 @@ ${JBANG_CMD} io.hyperfoil.tools:qDup:0.11.0 \
     -S config.jvm.version=${JAVA_VERSION} \
     -S config.quarkus.native_build_options="${NATIVE_QUARKUS_BUILD_OPTIONS}" \
     -S config.jvm.args="${JVM_ARGS}" \
+    -S config.energy=${ENERGY} \
```
edeandrea: Do you think there will ever be more "stuff" in `energy` where having a `config.energy` top-level object with other things inside would be beneficial?

scripts/perf-lab/main.yml
```diff
@@ -448,164 +404,195 @@ scripts:
...
+        - queue-download: ${{REPO_DIR}}/logs/measure-rss-${{RUNTIME.name}}-${{ITERATION}}.log
         - script: start-test-services
         - script: sync-drop-fs-cache
-        - sh: ${{RUNTIMECMD.runCmd}} &>${{REPO_DIR}}/logs/measure-rss-${{RUNTIMECMD.name}}-${{ITERATION}}.log &
+        - read-state: ${{config.energy}}
+          then:
+            - regex: "rapl|all"
+              then:
+                - sh: "[[ -x ${{RUN.RAPL_BIN:__NORAPL__}} ]] && echo RAPL_READY"
+                - regex: RAPL_READY
```
edeandrea: there should be a `then` block under `sh`. This is a breaking change with qDup 0.11 - see https://github.com/Hyperfoil/qDup/releases/tag/0.11.0

scripts/perf-lab/main.yml
```diff
@@ -448,164 +404,195 @@ scripts:
...
+                  then:
+                    - sh: sudo ${{RUN.RAPL_BIN}} -p > ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl &
+                    - sh: echo $!
+        - read-state: ${{config.energy}}
+          then:
+            - regex: "idrac|all"
+              then:
+                - sh: "command -v racadm > /dev/null 2>&1 && echo IDRAC_READY"
```
edeandrea: there should be a `then` block under `sh`. This is a breaking change with qDup 0.11 - see https://github.com/Hyperfoil/qDup/releases/tag/0.11.0

scripts/perf-lab/main.yml
```diff
@@ -448,164 +404,195 @@ scripts:
...
         - script: state-array-push
           with:
-            array: RUN.output.results.${{RUNTIMECMD.name}}.rss.firstRequest
+            array: RUN.output.results.${{RUNTIME.name}}.rss.firstRequest
             value: ${{= ${{FIRST_REQUEST}}/1024 }}
+        - read-state: ${{config.energy}}
+          then:
+            - regex: "idrac|all"
+              then:
+                - sh: "command -v racadm > /dev/null 2>&1 && echo IDRAC_READY"
```
edeandrea: there should be a `then` block under `sh`. This is a breaking change with qDup 0.11 - see https://github.com/Hyperfoil/qDup/releases/tag/0.11.0

scripts/perf-lab/main.yml
```diff
@@ -448,164 +404,195 @@ scripts:
...
+                - regex: IDRAC_READY
+                  then:
+                    - signal: ${{COMPLETE_SIGNAL}}
+                    - sh: echo "${{RUN.svrPwr}}" > ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.svrPwr
+                    - queue-download: ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.svrPwr
```
edeandrea: I might add the `queue-download` things further up in the script, that way if something dies before it gets here, the log file will get downloaded.

As written here, the log file will never get downloaded if there is a failure prior to this step.

scripts/perf-lab/main.yml
```diff
@@ -448,164 +404,195 @@ scripts:
...
+        - read-state: ${{config.energy}}
+          then:
+            - regex: "rapl|all"
+              then:
+                - sh: "[[ -x ${{RUN.RAPL_BIN:__NORAPL__}} ]] && echo RAPL_READY"
```
edeandrea: there should be a `then` block under `sh`. This is a breaking change with qDup 0.11 - see https://github.com/Hyperfoil/qDup/releases/tag/0.11.0

scripts/perf-lab/helpers/energy.yml
```diff
@@ -0,0 +1,39 @@
...
+                - script: state-array-push
+              with:
+                array: RUN.svrPwr
+                value: ${{watts}}
+  install-rapl-plot:
+    - read-state: ${{config.energy}}
+      then:
+        - regex: "rapl|all"
+          then:
+            - sh: "[[ -d /sys/class/powercap/intel-rapl ]] && echo RAPL_SUPPORTED || echo RAPL_UNSUPPORTED"
```
edeandrea, line 136: there should be a `then` block under `sh`. This is a breaking change with qDup 0.11 - see https://github.com/Hyperfoil/qDup/releases/tag/0.11.0

scripts/perf-lab/helpers/energy.yml
```diff
@@ -0,0 +1,39 @@
...
+            - regex: RAPL_SUPPORTED
+              then:
+                - sh: "[[ -f ${{RAPL_PATH:/tmp/rapl}}/${{RAPL_EXE:rapl-plot}} ]] || echo GETRAPL;"
```
edeandrea: there should be a `then` block under `sh`. This is a breaking change with qDup 0.11 - see https://github.com/Hyperfoil/qDup/releases/tag/0.11.0

scripts/perf-lab/main.yml
```diff
@@ -448,164 +404,195 @@ scripts:
...
         - script: sync-drop-fs-cache
-        - sh: ${{RUNTIMECMD.runCmd}} &>${{REPO_DIR}}/logs/measure-rss-${{RUNTIMECMD.name}}-${{ITERATION}}.log &
+        - read-state: ${{config.energy}}
+          then:
+            - regex: "rapl|all"
+              then:
+                - sh: "[[ -x ${{RUN.RAPL_BIN:__NORAPL__}} ]] && echo RAPL_READY"
+                - regex: RAPL_READY
+                  then:
+                    - sh: sudo ${{RUN.RAPL_BIN}} -p > ${{OUTPUT_DIR}}/${{COMPLETE_SIGNAL}}.rapl &
```
edeandrea: We should call the `sudo` script rather than run `sh: sudo ...`

scripts/perf-lab/helpers/energy.yml
```diff
@@ -0,0 +1,39 @@
...
+    - set-state: RUN.svrPwr []
+    - repeat-until: ${{COMPLETE_SIGNAL}}
+      then:
+        - sleep: 1s
+        - sh: racadm get system.Power.Realtime.Power
+          then:
+            - regex: (?<watts>\\d+) W | (?<btu>\\d+) Btu/hr
+              then:
+                - script: state-array-push
+              with:
```
edeandrea, line 74: the `with` should be aligned with `script`

```yaml
- script: state-array-push
  with:
```

scripts/perf-lab/run-benchmarks.sh
holly-cummins: Yes, good point. I think it's enough of a possibility we should. It could possibly be something like `energy:  {idrac: enabled, rapl: enabled}` and then if we get more ways of measuring, like scaphandre on mac, it doesn't mess up the enumeration?

scripts/perf-lab/helpers/energy.yml
holly-cummins: If you can't trust an LLM to format things correctly, what *can* you trust? :)

scripts/perf-lab/run-benchmarks.sh
edeandrea: That looks good to me. Just thinking about the future and not having to break things.

scripts/perf-lab/run-benchmarks.sh
```diff
@@ -37,6 +37,12 @@ help() {
   echo "  --cpus-otel <CPUS_OTEL>                                 CPU list for the OpenTelemetry stack (e.g. 14,16,18)"
   echo "                                                              Default: ${CPUS_OTEL}"
   echo "  --description <DESCRIPTION>                             A human-readable description to be added to the run output"
+  echo "  --energy-idrac <ENERGY_IDRAC>                             Enable iDRAC (system-level) energy measurement"
```
edeandrea: I think you can still leave it as a single `--energy` flag, where the value could be multiples?

`--energy idrac,rapl` or `--energy all`

Similar to how we do the runtimes/tests

holly-cummins: Oh, yes, that'd be cleaner.

scripts/perf-lab/main.yml
```diff
@@ -448,164 +408,246 @@ scripts:
...
         - script: stop-test-services
     - script: state-array-calc-avg
       with:
-        var-name: RUN.output.results.${{RUNTIMECMD.name}}.startup.avStartTime
-        array: RUN.output.results.${{RUNTIMECMD.name}}.startup.timings
+        var-name: RUN.output.results.${{RUNTIME.name}}.startup.avStartTime
+        array: RUN.output.results.${{RUNTIME.name}}.startup.timings
 
-  measure-rss:
+  measure-rss: # also measures energy
```
edeandrea: So if we're "tacking" energy management onto the `measure-rss` test, if the user specifies `--energy all` but does not specify the `measure-rss` test to run, the energy tests won't run.

We can solve that 2 ways:
1. Create a separate `measure-energy` test that the user would have to specify to run
2. In `run-benchmarks.sh`, if the user specifies `--energy` but does not specify `measure-rss` in the tests to run, add `measure-rss` to the set of tests to run.

I'd vote for option 2. Option 1 is somewhat cleaner, but doesn't totally solve the problem. You'd still need option 2 + s/measure-rss/measure-energy/

We're already doing something similar in `run-benchmarks.sh`:

```bash
  # Strip measure-build-times from TESTS_TO_RUN - it always runs during the build phase
  filtered_tests=()
  for item in ${TESTS_TO_RUN[@]}; do
    if [[ "$item" != "measure-build-times" ]]; then
      filtered_tests+=("$item")
    fi
  done
  TESTS_TO_RUN=${filtered_tests[@]}
```

You could add some logic there

holly-cummins: I was coming here to say option 2 as well. Will update!

scripts/perf-lab/run-benchmarks.sh
```diff
@@ -576,14 +607,12 @@ if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
...
   done
 
-  # Strip measure-build-times from TESTS_TO_RUN - it always runs during the build phase
-  filtered_tests=()
-  for item in ${TESTS_TO_RUN[@]}; do
-    if [[ "$item" != "measure-build-times" ]]; then
-      filtered_tests+=("$item")
+  # If user explicitly set --energy, ensure measure-rss is in the test list
+  if [[ -n "${ENERGY_SET_BY_USER:-}" && "$ENERGY" != "none" ]]; then
+    if [[ ! "${TESTS_TO_RUN[@]}" =~ "measure-rss" ]]; then
```
edeandrea: Couldn't this be a single `if` statement?

`if [[ -n "${ENERGY_SET_BY_USER:-}" && "$ENERGY" != "none" &&  ! "${TESTS_TO_RUN[@]}" =~ "measure-rss" ]]; then`

holly-cummins: I agree, that's nicer. Done.

edeandrea:
I ran this locally. It looks like rapl worked, but the idrac didn't.

```
14:47:03,809 [ test-runner:run-tests-1354 @ target-host ] [[ -x /tmp/rapl/rapl-plot ]] && echo RAPL_READY
RAPL_READY
14:47:03,810 [ test-runner:run-tests-1354 @ target-host ] regex: RAPL_READY
14:47:03,810 [ test-runner:run-tests-1354 @ target-host ] script-cmd: sudo
14:47:03,810 [ test-runner:run-tests-1354 @ target-host ] sudo
14:47:03,811 [ test-runner:run-tests-1354 @ target-host ] script-cmd: sudo-${{os}}
14:47:03,811 [ test-runner:run-tests-1354 @ target-host ] sudo-linux
14:47:04,236 [ test-runner:run-tests-1354 @ target-host ] sudo /tmp/rapl/rapl-plot -p > /home/hyperfoil/spring-quarkus-perf-comparison/logs/measure-rss-quarkus3-jvm-0.rapl &
[1] 1800810
14:47:04,661 [ test-runner:run-tests-1354 @ target-host ] echo $!
1800810
14:47:04,661 [ test-runner:run-tests-1354 @ target-host ] set-state: RUN.RAPL_PID 1800810
14:47:04,661 [ test-runner:run-tests-1354 @ target-host ] read-state: enabled
14:47:04,661 [ test-runner:run-tests-1354 @ target-host ] regex: enabled
14:47:05,091 [ test-runner:run-tests-1354 @ target-host ] command -v racadm > /dev/null 2>&1 && echo IDRAC_READY
14:47:05,520 [ test-runner:run-tests-1354 @ target-host ] taskset --cpu-list 0-3 java -Xms512m -Xmx512m -XX:+UseNUMA  -jar /home/hyperfoil/spring-quarkus-perf-comparison/spring-quarkus-perf-comparison/builds/quarkus3-jvm/quarkus-app/quarkus-run.jar &>/home/hyperfoil/spring-quarkus-perf-comparison/logs/measure-rss-quarkus3-jvm-0.log &
```

i never see the regex check for `IDRAC_READY`

I think the problem is with `- sh: "command -v racadm > /dev/null 2>&1 && echo IDRAC_READY"`

The echo never happens because the console output is redirected to `/dev/null`

Also, whats the purpose of the `sh` followed by the `regex`? shouldn't it always echo `IDRAC_READY`? and if so, why use regex? Could it ever be anything else?

It also looks like its running that same command in multiple places?

Additionally,

```yaml
  capture-system-power:
    - script: idrac-login
      with:
        idrac: ${{= ${{servers}}.filter( server => { return server.hostname == "${{env.HOST}}"})[0].idrac}}
        password: ${{idrac_pwd}}
```

where is `idrac_pwd` ever set? or `servers` for that matter? Furthermore, `capture-system-power` seems to be calling another script, `idrac-login`, but that script doesn't seem to exist.

edeandrea:
This may just be coincidence, but I ran both `quarkus3-jvm` and `spring4-jvm`, and the results for `energy` were the same for both.

```json
"energy": {
        "raplPkg": [
          0.44
        ],
        "raplDram": [
          0.53
        ],
        "avRaplPkg": 0.44,
        "avRaplDram": 0.53
      }
```

full results:

```json
{
  "timing": {
    "start": "2026-04-17T18:45:16Z",
    "stop": "2026-04-17T18:50:15Z"
  },
  "results": {
    "quarkus3-jvm": {
      "build": {
        "timings": [
          9.21
        ],
        "avBuildTime": 9.21
      },
      "startup": {
        "timings": [
          2494.544875
        ],
        "avStartTime": 2494.544875
      },
      "rss": {
        "startup": [
          270.3828125
        ],
        "firstRequest": [
          288.33984375
        ],
        "avStartupRss": 270.3828125,
        "avFirstRequestRss": 288.33984375
      },
      "energy": {
        "raplPkg": [
          0.44
        ],
        "raplDram": [
          0.53
        ],
        "avRaplPkg": 0.44,
        "avRaplDram": 0.53
      }
    },
    "spring4-jvm": {
      "build": {
        "timings": [
          3.91
        ],
        "avBuildTime": 3.91
      },
      "startup": {
        "timings": [
          6873.358205
        ],
        "avStartTime": 6873.358205
      },
      "rss": {
        "startup": [
          610.58203125
        ],
        "firstRequest": [
          636.1015625
        ],
        "avStartupRss": 610.58203125,
        "avFirstRequestRss": 636.1015625
      },
      "energy": {
        "raplPkg": [
          0.44
        ],
        "raplDram": [
          0.53
        ],
        "avRaplPkg": 0.44,
        "avRaplDram": 0.53
      }
    }
  },
  "config": {
    "jvm": {
      "args": "-XX:+UseNUMA",
      "memory": "-Xms512m -Xmx512m",
      "graalvm": {
        "version": "25.0.2-graalce",
        "home": ""
      },
      "version": "25.0.2-tem",
      "home": ""
    },
    "num_iterations": 1,
    "quarkus": {
      "build_config_args": "",
      "native_build_options": "",
      "version": "3.34.3"
    },
    "repo": {
      "short_commit": "60f1a44",
      "scenario": "tuned",
      "commit": "60f1a446fd208a2fdf7916f6bb716274227a8fb0",
      "branch": "lab-energy-measurement",
      "scenarioName": "Tuned",
      "url": "https://github.com/holly-cummins/spring-quarkus-perf-comparison"
    },
    "springboot3": {
      "native_build_options": "",
      "version": "3.5.13"
    },
    "springboot4": {
      "native_build_options": "",
      "version": "4.0.5"
    },
    "resources": {
      "cpu": {
        "app": "0-3",
        "1st_request": 10,
        "otel": "7-9",
        "monitor": 13,
        "load_generator": "10-12",
        "db": "4-6"
      },
      "app_cpus": 4
    },
    "run": {
      "dropOsFilesystemCaches": "true",
      "description": "energy",
      "useContainerHostNetwork": "true"
    },
    "units": {
      "rss": {
        "load": "MiB",
        "startup": "MiB",
        "firstRequest": "MiB"
      },
      "load": {
        "throughputDensity": "tps per MiB",
        "throughput": "tps",
        "errors": {
          "connectionErrors": "absolute number",
          "requestTimeouts": "absolute number"
        }
      },
      "timings": {
        "build": "sec",
        "startup": "ms"
      },
      "energy": {
        "raplDram": "W",
        "raplPkg": "W",
        "idrac": "W"
      }
    },
    "profiler": {
      "name": "none",
      "events": "cpu"
    },
    "energy": {
      "rapl": "enabled",
      "idrac": "enabled"
    }
  },
  "env": {
    "host": {
      "memory": "62Gi",
      "os": "Fedora Linux 43 (Workstation Edition)",
      "kernel": "6.19.10-200.fc43.x86_64",
      "cpu": "Intel(R) Core(TM) Ultra 7 165H (22 cpus)",
      "type": "LENOVO 21KWS49V00",
      "gpu": "Intel Corporation Meteor Lake-P [Intel Arc Graphics]"
    }
  }
}

```

franz1981:
As pointed by @edeandrea maybe something is off, but I still think we should compute the Joule from the energy drawn (that means that the time of startup matter) - and eventually move to https://greensoftware.foundation/articles/gps-up-a-better-metric-for-comparing-software-energy-efficiency/ too

holly-cummins:
> never see the regex check for `IDRAC_READY`
> 
> I think the problem is with `- sh: "command -v racadm > /dev/null 2>&1 && echo IDRAC_READY"`
> 
> The echo never happens because the console output is redirected to `/dev/null`

Claude says 
> The redirect is confusing (though it's actually correct — >/dev/null only applies to command -v, not the echo after &&)                        
 
> Also, whats the purpose of the `sh` followed by the `regex`? shouldn't it always echo `IDRAC_READY`? and if so, why use regex? Could it ever be anything else?
> 
> It also looks like its running that same command in multiple places?

It agrees with you. :) 

> 2. The sh+regex guard is roundabout — it's a runtime check for racadm availability, repeated identically each time                                
>  3. It's duplicated 3 times in measure-rss                                                                                                         
>                                                                                                                                                    
>  The fix: check once during setup, store the result in state, and use read-state guards everywhere else. This mirrors how RAPL is validated during install-rapl-plot.        

I've fixed those points. I've also put in a best guess at a `idrac-login` script, since I can't find a model one anywhere in the original scripts. There is a default password but the systems push users to change them, so we have to assume it's something else, and configure that.

holly-cummins:
> As pointed by @edeandrea maybe something is off, but I still think we should compute the Joule from the energy drawn (that means that the time of startup matter) - and eventually move to https://greensoftware.foundation/articles/gps-up-a-better-metric-for-comparing-software-energy-efficiency/ too

You're totally right. I'm a bit sceptical about making GPS-UP our primary metric because we want to be able to communicate to people without them having to go read a document. I think it's maybe a bit too specialised for this use case, too – the question we're asking isn't "I want to evaluate an optimisation," but more "I want to know what my energy bill might be."

But I'd got muddled about what the RSS test was doing, which may be partly why @edeandrea's results make no sense. We *know*, from both local measurement and our earlier publications, there's an energy difference. I'd imagined the RSS test was running for a fixed time, and taking a mean or max RSS after a warmup period. If that was the case, average and total energies would be strictly proportional, and it wouldn't really matter which I measured. (Technically, the average would be a power, not an energy, but that's a detail. :) )

But on looking more closely, I see that the RSS test only runs until the first request comes back. So that means 
(a) the test is so short that it will be hard to measure anything meaningfully. At one sample a second, the noise will be huge.
(b) average is totally wrong as a metric because it would give no benefit to starting rapidly

So we could add a whole other test, or we could extend the RSS test if energy is enabled. I'll go for the 're-use but extend' option, and switch to measuring total.

scripts/perf-lab/helpers/energy.yml
```diff
@@ -0,0 +1,71 @@
+name: Energy-related scripts
+scripts:
+  capture-system-power:
+    - script: idrac-login
+      with:
+        idrac: ${{= ${{servers}}.filter( server => { return server.hostname == "${{env.HOST}}"})[0].idrac}}
+        password: ${{idrac_pwd}}
```
edeandrea: Where are the variables `servers` and `idrac_pwd` set? I don't see them anywhere?

holly-cummins:
I kept looking for this comment from @zakkak, thinking it was on this PR, but it's elsewhere. Pasting it here since it's relevant (but probably an advanced iteration of the basics done here): https://github.com/quarkusio/spring-quarkus-perf-comparison/issues/387#issuecomment-4279025625

scripts/perf-lab/helpers/energy.yml
holly-cummins: Claude says, on my behalf [without asking, I note]:

Good catch — they were never wired up. I've simplified this: since the benchmarks run on the Dell server itself, `racadm` can talk to the local iDRAC directly without any host, username, or password. Removed the `servers` array lookup and `idrac_pwd` entirely. The `idrac-login` script now just validates that local `racadm getsysinfo` works.

holly-cummins: Ugh, another good catch. I kind of think we don't have enough information (like passwords!) to make idrac work and we should be happy if rapl works, but if idrac is in the codebase, it needs a chance of working. 

My friend Claude, the brains of this whole operation, says 
>  The reviewer is right — servers and idrac_pwd are never set anywhere. The servers array with a JS filter is also overengineered for a single-host 
  setup. I'll simplify by replacing it with a direct idrac_host variable and wire both through run-benchmarks.sh.                                   

On further questioning, I also got Claude to admit that if the scripts run on the same machine as the job execution, idrac credentials aren't even needed. So it ends up much simpler now, and maybe has a chance of working on Dell.

edeandrea: Does it "only" work on dell machines?

holly-cummins: I read this as 

> Does "only" work on dell machines?

and I was *seriously* impressed at the level of the review. "Wow, Eric is picking up vendor-level differences in shell script behaviour". 

Then I read properly. :) 

Yes, as far as I know, idrac is Dell-only. It's similar to RAPL but does a broader measurement, to include things like fan-power. I included it because John did, but from my perspective, I'm assuming the 'overall system usage' that idrac gives us is a multiple of what RAPL would give us, and since we're interested in the ratio between frameworks, we probably don't care all that much.

edeandrea:
FYI @holly-cummins I re-ran this and I'm still getting the exact same values for both Quarkus & Spring

```json
      "energy": {
        "raplPkg": [
          27
        ],
        "raplDram": [
          33
        ],
        "avRaplPkg": 27,
        "avRaplDram": 33
      }
```

scripts/perf-lab/helpers/energy.yml
edeandrea: > and I was seriously impressed at the level of the review. "Wow, Eric is picking up vendor-level differences in shell script behaviour".

Well, you said it first :)

`...and maybe has a chance of working on Dell.`

And I also tried running `racadm` on my lenovo and it wasn't there :)

edeandrea:
Is the `measure-rss` test the right test for this? Like you said, its a quick test - start the app (measure rss), do a single `curl` (measure rss).

Has enough happened to know how much energy was consumed? Shouldn't energy measurement have a time component to it? All of the tests in `measure-rss` run for a fixed amount of time. App start command is issued, the test pauses for a set duration of time, then issues a `curl`, then shuts the app down.

Or should energy measurement be a component of each test? (Time to first request, load test, rss, build time, etc)? I would think that energy expended during load testing could be interesting. If an app can handle lots of request, but we find that the app threads are "asleep" waiting on cpu, maybe the app uses less energy?

I'm also not sure what `- sh: sleep ${{config.energy.rss.duration}}` is doing?

holly-cummins:
> Is the `measure-rss` test the right test for this? Like you said, its a quick test - start the app (measure rss), do a single `curl` (measure rss).
> 
> Has enough happened to know how much energy was consumed? Shouldn't energy measurement have a time component to it? 

Yes, definitely. It's *supposed* to be putting load on the system for 150s, same duration as the normal tests. But I suppose it's possibly just letting the system run for 150s, with no load. That would be rather sad, but would explain the strange results. 

> All of the tests in `measure-rss` run for a fixed amount of time. App start command is issued, the test pauses for a set duration of time, then issues a `curl`, then shuts the app down.

Oh dear. 

> Or should energy measurement be a component of each test? (Time to first request, load test, rss, build time, etc)? I would think that energy expended during load testing could be interesting. If an app can handle lots of request, but we find that the app threads are "asleep" waiting on cpu, maybe the app uses less energy?

I think energy testing needs to be done with fixed load. Well, that's not quite true. It possibly would be interesting to report energy for all of the tests, because if we can handle 2x throughput with less energy, that's *very* interesting, but I think on the "vary one thing at a time" principle that @franz1981 has drilled into us, the 'best' energy measurement is done at fixed throughput, so it's an "all other things are equal" test. 

> I'm also not sure what `- sh: sleep ${{config.energy.rss.duration}}` is doing?

I kind of assumed qDup had two threads so that thread was just waiting for the load to execute. But maybe that's actually not at all what it's doing, and it's just wasting time. :(

edeandrea:
> Yes, definitely. It's supposed to be putting load on the system for 150s, same duration as the normal tests. But I suppose it's possibly just letting the system run for 150s, with no load. That would be rather said, but would explain the strange results.

There's no load on the system during that 150s. The curl has already happened. The app is just sitting there idle without anything hitting it.

scripts/perf-lab/helpers/energy.yml
holly-cummins: I dropped the 'it' in my head and thought it was a new cli tool, "only".

franz1981:
Interestingly, it's actually easier to get the max TPS test right than the fixed throughput one. By saturating the CPU, we're effectively running at a fixed *hardware* load (100% utilization) and just measuring the work done (Requests per Joule).

Not to mention, if we use a fixed TPS, we run into a sneaky issue: if one framework is much more efficient and uses far fewer resources, it will spend more time sleeping. In that scenario, the energy cost of constant CPU wakeups could actually end up being higher than the cost of handling the requests that originated them! 

So we have to think about this carefully before we commit entirely to fixed TPS for energy baselines.
In short let's experiment with both and analyze which work best and what they mean to users 🙏

I know that I am contradicting myself here, but for scientific reasons I cannot ignore the challenges of testing each scenario 🙏

holly-cummins:
> So we have to think about this carefully before we commit entirely to fixed TPS for energy baselines. In short let's experiment with both and analyze which work best and what they mean to users 🙏
> 
> I know that I am contradicting myself here, but for scientific reasons I cannot ignore the challenges of testing each scenario 🙏

Since I'm doing such an *awesome* job of getting one scenario right, I think doing two scenarios is an excellent idea. :) 

(Joking aside, I agree, let's experiment and think. In that order, probably. :) )

holly-cummins:
Another option for measuring is to ramp up the requests, the same as what we had in the original measurements:

image

That would be expensive to do every time, though.

edeandrea:
@holly-cummins the energy capture - is it looking at a specific process? Or is it looking at the system as a whole during a period of time?

If its capturing the total power consumption of the system, wouldn't that capture both the app under load and the load generator (plus database and other things running)?

zakkak:
> Not to mention, if we use a fixed TPS, we run into a sneaky issue: if one framework is much more efficient and uses far fewer resources, it will spend more time sleeping. In that scenario, the energy cost of constant CPU wakeups could actually end up being higher than the cost of handling the requests that originated them!

@franz1981, who would be triggering these "constant CPU wakeups" the framework itself or the energy measurement harness? If it's the framework itself then it sounds fair game to me. The CPU wakeups should be priced in...

> Another option for measuring is to ramp up the requests, the same as what we had in the original measurements:

@holly-cummins Tons is a bit scary on the y-axis :D

> That would be expensive to do every time, though.

+1, it's pretty nice though so it might be worth running on each new feature release (once a month?). How long does it take to run?

franz1981:
> who would be triggering these "constant CPU wakeups

The originating cause is: network interrupts due to new requests arriving (+ causal dependency between thread pools due to Quarkus threading model). But the actual motivation is because each of the involved thread pools (netty and any workers) could have already completed their job., they have an higher chance to go park - and be unparked.

The slower framework instead, is kept awakw...making it appears as efficient (or more?) depending by the machine settings for C state.

That's why is dangerous to measure just a single fixed TPS, but you would like to measure at different load - exactly like .measuring efficiency of cars under different scenarios (type of roads, different speeds, acceleration/deceleration, etc etc)

holly-cummins:
@edeandrea and I were talking the other day, and the way our test is setup may make energy measurement in the max load case non-interesting. I'd still like to collect it, but I don't think we can draw very many conclusions from it. Because RAPL collects whole-system power, and because the DBMS and load generator are co-located on the system, we'd end up measuring them as well. If their work is (relatively) constant, we can just treat it as a baseline in the comparison. But if Quarkus is able to handle more load, the load generator and DBMS will be doing more work, and consuming more power.  We wouldn't be able to draw conclusions about the efficiciency of the frameworks.

So it sounds like we'll want to be doing a few different loads as our 'default' measurement, although I'd also like to report a 'slice though' metric. In our old data, we did see a fairly linear relationship between load and energy over wide ranges

image

... but less linear in small ranges, which I guess would be consistent with @franz1981's concerns.

image

zakkak:
> The originating cause is: network interrupts due to new requests arriving (+ causal dependency between thread pools due to Quarkus threading model). But the actual motivation is because each of the involved thread pools (netty and any workers) could have already completed their job., they have an higher chance to go park - and be unparked.

Good, so it's part of the framework (or its underlying dependencies) in which case it's fair to measure it. No?

> The slower framework instead, is kept awakw...making it appears as efficient (or more?) depending by the machine settings for C state.

@franz1981 I am confused, why would an awake CPU be more efficient than a sleeping one that's being waken up?

> That's why is dangerous to measure just a single fixed TPS, but you would like to measure at different load - exactly like .measuring efficiency of cars under different scenarios (type of roads, different speeds, acceleration/deceleration, etc etc)

I totally agree on that.

> Because RAPL collects whole-system power, and because the DBMS and load generator are co-located on the system, we'd end up measuring them as well.

@holly-cummins is there no option to run the load generator and DBMS on a different system? It would sure add some overhead which would drop the peak throughput, but would make the results more realistic and representative IMHO (no matter if using a fixed throughput or going for peak). Keeping the DBMS in the mix might be OK, but the load generator doesn't make much sense.

franz1981:
@zakkak

> it's part of the framework (or its underlying dependencies) in which case it's fair to measure it. No?

It depends — it's more a timing artifact than a framework property. Ideally you'd measure each framework at *its own* 75% load (which differs between them), or size the more efficient one to match the peak throughput of the slower one, and then run both at the same fixed TPS. That makes the comparison meaningful in terms of capacity served.

> I am confused, why would an awake CPU be more efficient than a sleeping one that's being waken up?

It's not more efficient in steady state — but it *appears* so under fixed TPS measurement. With turboboost disabled and a stable P-state, a warm CPU draws a predictable, moderate amount of energy. A CPU that enters deep C-states between requests pays a wakeup tax on every request arrival: PLL relock, cache warm-up, a potential turbo burst. That cost gets attributed to the efficient framework precisely *because* it finishes fast enough to sleep. The slow framework never parks, avoids all that overhead, and ends up looking comparably efficient — or better — depending on the machine's C-state configuration.

zakkak:
> Depends. It's more an artifact of some bad timing ^^ you want to measure for both frameworks how they perform at THEIR 75% load/vs peak (which is different)

Hmm I find Energy per request more interesting/representative. Running at 75% capacity doesn't say much when the capacity is not the same.

Service providers probably have a rough estimate of how many requests they anticipate on a given period and are probably interested in minimizing the total cost of serving that requests while meeting some performance criteria. So I would expect them to be interested in `cost per request` which is tightly coupled with `energy per request` (when running on premise at least).

> - or size the most efficient one to serve the same peak of the slower - and than you can measure the same fixed one against both - which are more comparable in term of "capacity".

I think this is tricky as well. Since both frameworks run on the same machine the one stressing the machine will be consuming more power than the other. At the same time, the "throttled" framework despite consuming less energy overall for the same throughput, it will have a worse `energy per request` ratio if I am correctly reading Holly's plot in https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/473#issuecomment-4312102009 which indicates that `energy consumption/request` improves linearly with TPS. So the highest the TPS the more efficient a framework is (even when compared with itself at lower TPS).

> If a CPU is kept warm and P state with turboboost is disabled, it consume a certain amount of energy, but going in deep C state + unparking is like measuring how much fuel a car consume while stopping/starting - it rather different than at steady state.

Sure, but that's OK IMHO if energy measurement is done end-to-end. On the other side if energy measurement is done by periodically sampling power consumption then the non-steady state might indeed skew things...

In the case of end-to-end measurements though I would still expect the start/stop to result in better energy consumption in most cases, just like my car is still expected to burn less fuel overall if it turns itself off at the traffic lights.

To take a step back, what question are we trying to answer with these numbers?

franz1981:
@zakkak

> In the case of end-to-end measurements though I would still expect the start/stop to result in better energy consumption in most cases, just like my car is still expected to burn less fuel overall if it turns itself off at the traffic lights.

Traffic lights are periodic and predictable — HTTP load isn't. wrk2 generates Poisson arrivals: inter-arrival times are exponentially distributed (CV=1), so even at a fixed mean rate you get bursts and gaps that the Linux idle governor (`menu`/`teo`) cannot reliably predict.

Each C-state has a **target residency** — a minimum dwell time for the sleep savings to outweigh the entry/exit overhead (C6: ~100–300µs). The governor uses timer deadlines and recent idle history to pick the right state, but under high-variance arrivals misprediction rate increases: it may commit to a deeper state and get pulled out before residency, paying the transition cost with reduced savings.

This doesn't necessarily flip the result — race-to-idle is still the dominant design philosophy of modern CPUs, and a busy framework burning dynamic power continuously isn't free either. But it does mean the efficiency gap measured at fixed TPS can be smaller than the raw throughput ratio suggests, and the magnitude is hardware, kernel-tuning, and load-point dependent.

That's why measuring at a single fixed TPS is risky — the result tells you as much about C-state behavior on that specific machine as it does about the frameworks.

> To take a step back, what question are we trying to answer with these numbers?

+💯

scripts/perf-lab/main.yml
```diff
@@ -58,7 +68,7 @@ states:
   PAUSE_TIME: 5
   PROFILER_JVM_ARGS:
   BASE_JAVA_CMD: ${{APP_CMD_PREFIX}} java ${{config.jvm.memory}} ${{config.jvm.args}} ${{PROFILER_JVM_ARGS}}
-  TESTS : [measure-time-to-first-request, measure-rss, run-load-test]
+  TESTS : [measure-time-to-first-request, measure-rss, run-load-test, measure-energy-fixed-throughput]
```
diegolovison, line 71: I don't believe a dedicated test is necessary. We can add a configuration variable to toggle energy measurement for any existing test instead.

Proposed Approach:

- Toggle via Variable: Use a flag like config.enable.energy.consumption.
- Trigger the Script: In existing tests (e.g., measure-time-to-first-request, measure-rss, run-load-test), call the energy measurement script just before the warm-up or steady-state phases.

Even if a custom test is required later, we should keep this same implementation logic.

scripts/perf-lab/main.yml
```diff
@@ -628,6 +642,7 @@ scripts:
           then:
             - abort: unable to connect to target application
         - sleep: 30s # Wait for the app to complete warmup
+        - script: start-energy-capture
```
diegolovison, line 645: Because the energy consumption becomes a metric and not a test. See my previous comment. We can now have a special QDup file for each kind of energy consumption measurement.

diegolovison (commented):
I think we should separate the metric from the test.

scripts/perf-lab/main.yml
holly-cummins: I did apply this approach originally, but the problem is that we don't have an existing test that runs for a long period at fixed load.

scripts/perf-lab/main.yml
holly-cummins: I do like the approach of energy being a metric, and I'm hoping we'll capture it in several of our tests (even if we decide the results are most useful for one or two).

scripts/perf-lab/main.yml
holly-cummins: What if we rename the test, to drop 'energy'? That keeps the separation between metric and test, but gives us the new test parameters we need.

franz1981: In any case since we cannot distinguish the cost of the different components this will make energy comparison across different tests not possible

holly-cummins:
> This doesn't necessarily flip the result — race-to-idle is still the dominant design philosophy of modern CPUs, and a busy framework burning dynamic power continuously isn't free either. But it does mean the efficiency gap measured at fixed TPS can be smaller than the raw throughput ratio suggests, and the magnitude is hardware, kernel-tuning, and load-point dependent.
> 
> That's why measuring at a single fixed TPS is risky — the result tells you as much about C-state behavior on that specific machine as it does about the frameworks.

I've been thinking about this more, and I think the 'gold standard' test would be a scalability test (like the one @franz1981 is adding in https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/492). But I think we'll probably also want to report a 'slice' through that scale, probably somewhere around the middle. 

> > To take a step back, what question are we trying to answer with these numbers?
> 
> +💯

It's a good question. I think the question, ultimately, is "how will moving to Quarkus help me achieve my sustainability goals by reducing my carbon footprint?" The answer there is two parts, with one being more complex than the other. The more complex answer is that higher transaction rates for the same hardware, combined with lower memory requirements, mean workloads can run on smaller hardware and have less embodied carbon. That ties closely into the "what container would we need?" and cost calculation efforts. The second answer is "what would my energy usage per transaction be?" I think to be fair, and measuring-only-one-thing-at-a-time, that should be measured at a fixed transaction rate. The question isn't "if I had twice as many customers once I switched to Quarkus, what would my energy cost per transaction be?". (I mean, if it was a realistic question, that would be kind of awesome, but I don't think "doubles your customer base overnight" is part of our value proposition.)

If modern processor architecture means Quarkus still consumes similar-ish energy even when it seems to be running at half its potential max capacity, that's useful information. It's not the answer we want, but it's going to more accurately reflect user interests than a theoretical energy usage that relies on throughput being higher to avoid scheduler/hardware penalties. 

As an aside, if we assume our results are still mostly-like our results the last time we measured energy, a vertical slice does show the energy savings ratio is less than the max-throughput-ratio, but maybe not hugely. How much depends on the baseline (these charts have a non-zero baseline). For JVM, it's the ratio between the red line and the purple line:

image

scripts/perf-lab/main.yml
diegolovison: > I did apply this approach originally, but the problem is that we don't have an existing test that runs for a long period at a fixed load.

I think we should have two separate PRs:
1) Create a test that represents: runs for a long period at a fixed load. (soak test)
2) Another PR that adds energy consumption metrics to any test

This is what we are doing downstream. Energy measurement is a metric that can be attached to any test.

holly-cummins: Before we refactor this PR into two separate PRs, could we confirm if it works? We know that we can't make any kind of evaluation of (2) without (1). I could do a (1) and merge it and then we do (2), but we might then need to revise (1). Since unpicking apart this PR into two has some complexity, I'd prefer to ensure we're unpicking the right things apart.

holly-cummins: I'm anxious to know if the scripts work since my first four attempts at the script most definitely did *not* work, so it seems unlikely the current attempt does work. :)

holly-cummins:
@zakkak asked where the rest of the plot from the preceding comment was, so here's a version which hopefully is clearer: 

image

PR #479, opened 2026-04-17
Configure Vert.x to use a single event loop for Loom integration with Quarkus

edeandrea:
Change suggested by @franz1981 

Maybe he can explain exactly why this is better when using loom?

### DIFF
--- a/quarkus3-virtual/src/main/resources/application.yml
+++ b/quarkus3-virtual/src/main/resources/application.yml
@@ -20,6 +20,9 @@ quarkus:
       sampler:
         ~: traceidratio
         arg: 0.1
+  vertx:
+    # with loom, 1 event loop is like a poller managing virtual threads
+    event-loops-pool-size: 1
 
 "%prod":
   quarkus:
--- a/quarkus3-virtual/src/main/resources/application.yml
+++ b/quarkus3-virtual/src/main/resources/application.yml
@@ -22,6 +22,7 @@ quarkus:
         arg: 0.1
   vertx:
     # with loom, 1 event loop is like a poller managing virtual threads
+    # See https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/479
     event-loops-pool-size: 1
 
 "%prod":


franz1981:
Yep; since we have quite a lot of context switching between 4 event loops and the fork join workers, including awaking/parking them - maybe having just one would save the cpu time spent by the OS to interleave them and to awake them.
And FYI Spring, by default run with a single tomcat poller, so it "kind of" align to it too, although, to be fair,  such poller doesn't handle the HTTP protocol encoding/decoding, while Netty does it.

We clearly need to try it in the lab too.

edeandrea:
@franz1981 @holly-cummins this is what I see when I run with this setting:

## Comparison: New Results vs [Current Results](https://github.com/quarkusio/benchmarks/blob/main/results/spring-quarkus-perf-comparison-latest-tuned.json)                      
                                                                                                                                                                                   
  ### `quarkus3-jvm`                                                                                                                                                               
                                                                                                                                                                                   
  | Metric | Current | New Results | Change |                                                                                                                                      
  |---|---|---|---|                                                                                                                                                                
  | avThroughput (tps) | 13,264.69 | 13,312.31 | **+0.36%** |                                                                                                                      
  | maxThroughputDensity (tps/MiB) | 17.0672 | 17.1114 | **+0.26%** |                                                                                                              
                                                                                                                                                                                   
  ### `quarkus3-virtual`                                                                                                                                                           
                                                                                                                                                                                   
  | Metric | Current | New Results | Change |           
  |---|---|---|---|
  | avThroughput (tps) | 12,240.92 | 13,748.35 | **+12.32%** |
  | maxThroughputDensity (tps/MiB) | 16.5339 | 20.0375 | **+21.19%** |                                                                                                             
                                                                                                                                                                                   
  ### Summary                                                                                                                                                                      
                                                                                                                                                                                   
  `quarkus3-jvm` is essentially flat (within noise). `quarkus3-virtual` shows a significant improvement — ~12% higher throughput and ~21% better throughput density in the new results.

holly-cummins:
Woo! That's really good.

holly-cummins:
Should we document that suggestion? Or open a conversation about defaults or auto-tuning, because of course no one reads the documentation. 

I guess we need to explore the space of when this *isn't* a good idea. Presumably that space is larger for non-Loom scenarios than Loom ones, but even with Loom there would be some topologies or workload shapes where it doesn't make sense.

edeandrea:
@holly-cummins @franz1981 Do you think this is ready to merge?

franz1981:
Let me give it a shot on RHEL 10 and my local machine too

franz1981 (approved):
ready to go for me!

holly-cummins:
We should do a lab run to get new numbers for this, @edeandrea. What's our plan for the next step for this? Blog article with a before-and-after? Guide update to recommend the tuning flag? Change the defaults in a few months? Do you have any thoughts, @cescoffier and @jponge?

franz1981:
@holly-cummins you mean another one for the latest changes w ParallelGC too? https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/479#issuecomment-4347478461 didn't had such.

I can help dissect and explain why it has helped ;)

edeandrea:
> We should do a lab run to get new numbers for this, @edeandrea. What's our plan for the next step for this? Blog article with a before-and-after? Guide update to recommend the tuning flag? Change the defaults in a few months? Do you have any thoughts, @cescoffier and @jponge?

I was waiting for #529 to be merged and then do a run after.

holly-cummins:
> @holly-cummins you mean another one for the latest changes w ParallelGC too? [#479 (comment)](https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/479#issuecomment-4347478461) didn't had such.
> 
> I can help dissect and explain why it has helped ;)

My thinking for this is "if Quarkus + Loom performs better when Vert.x is configured to use a single event loop, then it's no use doing that change in the benchmark and forgetting about it. If everyone else should be using the same setting, we should do something about that."

And actually, if that is the case, maybe we'd want to do a lab run without the GC changes? Otherwise we don't have a set of published results we can point to to show the 'before' and 'after' in the docs and /or blog post.

franz1981:
Eric has run it without GC changes the first time, not sure he posted the results here...
Anyway is very specific of this workload and the limiting factor:
- netty event loops threads are way too many for the http activity: if you keep a single one a bit more busy it bother less the OS with many context switches because they exists, and they need to go to sleep If they got nothing to do
- If a single one is busier, the Quarkus arch requires to still ask Netty to send the response and awake some event loop, but being one and more awake, this cost is lessen, improving the VirtualThread carriers to save some wakeup to Netty

And, Spring use a single Poller in tomcat, too.

In general, keep threads warm is a good idea if they are not overloaded; profiling data revealed that the amount of CPU work required by Netty is way less than one core, so it's fine to reduce them - given that the other threads (virtual ones) are CPU hungry, and the more they can have (or save) the better is

holly-cummins:
> Eric has run it without GC changes the first time, not sure he posted the results here... 

I know we know it improves things, but for a blog post, it's nice to have published results to point to. 

> Anyway is very specific of this workload and the limiting factor:

Ah, ok, that's kind of what I was trying to figure out – is this specific tuning for this benchmark, or something we should be sharing?

I kind of wonder if we should have an informal policy that any tuning we make should be backed up by a blog or equivalent, explaining to people under what circumstances they should make the same change. That's in line with our guiding principle that we'll only do tunings we expect everyone to do, because this is not TechEmpower. So any tuning that we do, we need to justify (a) why we're not changing the Quarkus default to match the new value, and (b) why everyone shouldn't be applying it to their applications. Obviously, tuning is workload-dependent, so (b) usually wouldn't be that hard, but we should be able to write it down.

franz1981:
> That's in line with our guiding principle that we'll only do tunings we expect everyone to do

+100

edeandrea:
> Eric has run it without GC changes the first time, not sure he posted the results here... 

See https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/479#issuecomment-4347478461 - those are the results with just this change (no GC change).

PR #493, opened 2026-04-22
Add --profiler-events and --profiler-args options to run-benchmarks.sh

### LINKED ISSUE #377: async profiler argumens should be customizable via bash script
Currently qDup allows to override/customize the profiler arguments and it would be great if we could do the same via the bash script, exposing it as a parameter. This allows to profile for different type of events or with specific profiler settings

franz1981:
Expose async-profiler event type and extra arguments as CLI options instead of hardcoding 'cpu'. This allows profiling for different event types (wall, lock, alloc) or passing extra asprof flags (--total, etc) without needing --extra-qdup-args.

Closes #377

This will prove useful to validate why, in the perf lab, Spring is suffering from a huge amount of non-voluntary context switch .e.g by using `-e context-switches` - see https://stackoverflow.com/a/64404504 (no need to specify the interval anymore, it's now 2 OOTB).
Just as a note for myself of the future: it was needed 2 because async-profiler's context switch event emission can generate an interrupt which can temporarly awake a parked OS thread, which will likely park again, right after, doubling effectively the number of context switches.

### DIFF
--- a/scripts/perf-lab/helpers/async-profiler.yml
+++ b/scripts/perf-lab/helpers/async-profiler.yml
@@ -40,12 +40,13 @@ scripts:
     - wait-for: ${{wait_start:}} # don't wait by default
     - sleep: ${{delay:0}}
     - sh: >
-        bin/asprof 
-        -d ${{duration:10}} 
-        -f ${{REPO_DIR}}/logs/${{format:flamegraph}}.${{suffix:html}} 
-        -o ${{format:flamegraph}} 
-        -e ${{events:cpu}} 
-        ${{="${{enableTotal : }}".toLowerCase() == "true" ? "--total" : ""}} 
-        --title ${{title:async-profiler}} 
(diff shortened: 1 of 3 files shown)


diegolovison:
If you tested, then we can merge.

It took me time as well to find out https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/493/changes#diff-26a21248a06906f0688cee5f9f81a203620975a28e21d421e8b7fedda09a0d29L292

edeandrea:
@franz1981 is this something ready to merge?

franz1981:
@edeandrea If this pass all the checks can be merged. It default to CPU events so we don't have to modify the Jenkins job and groovy file immediately

franz1981:
@edeandrea this was ready to go, I have to check why is out of sync

diegolovison:
@franz1981 I can take care of it if you wish

PR #495, opened 2026-04-22
feat(openapi): Add OpenAPI for the development

jabrena:


edeandrea:
I totally agree contact-first development is a good thing. Starting with a schema, then having your build process generate some interfaces that you then implement, is also good.

But in the context of these tests, do we _really_ want to do that? To me that means giving up control to some generator, which will be different for Quarkus vs spring (Quarkiverse has its own openapi generator for rest services, Spring i'm sure has something too). Who knows the quality of the code it generates and what overhead that code adds to the benchmarks.

I'm more apt to say "sure lets have an OpenAPI document for people to reference", but I'm hesitant to implement true contract-first for any of the apps in this repo.

@holly-cummins thoughts?

jabrena:
We could use it for black box testing

edeandrea:
Closing this in favor of #496

PR #498, opened 2026-04-23
Fix benchmark-orchestrator image build on linux/arm64

cdhermann:
JBang's curl installer downloads JDK 17 on its own which consistently fails on arm64 GitHub runners. Also adds tag sanitization to the workflow so branch names containing slashes produce valid Docker image tags.

- Add java-21-openjdk-devel to the dnf install step (UBI9 native package, works on both amd64 and arm64)
- Install JBang via SDKMAN instead of the curl installer — SDKMAN is already present, handles arm64 natively, no separate JDK download needed
- Sanitize github.ref_name before using it as a Docker image tag (slashes in branch names are invalid in image tag references)

### DIFF
--- a/.github/workflows/build-publish-qdup-ghcr.yml
+++ b/.github/workflows/build-publish-qdup-ghcr.yml
@@ -44,6 +44,10 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Sanitize tag
+        id: tag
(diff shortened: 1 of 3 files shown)


edeandrea:
So that benchmark orchestrator image - I never really spent any time trying to get it to work :)

cdhermann:
> So that benchmark orchestrator image - I never really spent any time trying to get it to work :)

I just compiles, I haven't tested it for being useful and handle stuff as expected

edeandrea:
@cdhermann I'm actually going to close this PR because I've never even verified that the orchestrator image even works. I attempted it a long time ago and gave up on it because running the perf test suite on a container image didn't seem like a good idea.

PR #512, opened 2026-04-27
Support native builds with Mandrel and fix load-test race conditions for extremely fast native startup

cdhermann:
### Overview

This PR adds support for building Quarkus native images with **Mandrel** alongside the existing GraalVM CE support. It also includes bug fixes for load-testing race conditions that were affecting benchmark measurement accuracy.

### Changes

- **9db0000**: Support native builds with Mandrel next to GraalVM
  - Adds Mandrel as an alternative native image compiler option
  - Maintains compatibility with existing GraalVM CE builds
  
- **49c1308**: Fix race condition in watch-log (async startup detection)
  - Native Quarkus binaries start in ~80ms, fast enough that the async watch-log script could fire `LOG_REGEX_REACHED` before the main thread registered its `wait-for`
  - Added 1s sleep between ctrlC and signal to ensure main thread registration

- **5b9739b**: Fix duplicate 'if' key in syncbot.yml (GitHub Actions validation)

- **da68703**: Fix load-test watch-log race (timing issue)
  - Moved watch-log invocation to **before** app start
  - Previously, tail -f would replay existing log content when the app had already started, causing regex to fire instantly and skewing results
  - Removed ineffective 1s sleep from watch-log callback

### Benchmark Findings

Related work on <https://github.com/cdhermann/spring-quarkus-perf-comparison/tree/native-mandrel-with-dotnet> branch includes:

- Demo benchmarking comparing GraalVM CE vs. Mandrel with .NET 10 integration
- **Key Finding**: GraalVM CE and Mandrel perform **equally well** in throughput and startup time metrics
- **Interesting Discovery**: Mandrel's reachable field, reflection types, reflection methods analysis detects **one less** compared to GraalVM CE; see <https://github.com/cdhermann/spring-quarkus-perf-comparison/blob/native-mandrel-with-dotnet/docs/benchmark-runs/results-quarkus-all-JDK25_and_dotnet10-7x-2026-04-27.md#native-image-statistics>

### Testing

- Verify native builds work with both GraalVM CE and Mandrel backends
- Run full load test suite to confirm race condition fixes don't impact measurements
- Validate log-based startup detection with fast-starting native binaries

### DIFF
--- a/.github/workflows/syncbot.yml
+++ b/.github/workflows/syncbot.yml
@@ -147,8 +147,7 @@ jobs:
           fi
 
       - name: Update PR comment on success
-        if: steps.push-and-pr.outputs.push_failed_workflows != 'true' && steps.cherry-pick.outputs.skipped_commits == ''
-        if: steps.cherry-pick.outputs.all_skipped != 'true' && steps.cherry-pick.outputs.skipped_commits == ''
+        if: steps.push-and-pr.outputs.push_failed_workflows != 'true' && steps.cherry-pick.outputs.all_skipped != 'true' && steps.cherry-pick.outputs.skipped_commits == ''
         uses: quarkusio/action-helpers@main
         with:
           action: maintain-one-comment
--- a/scripts/perf-lab/helpers/java.yml
+++ b/scripts/perf-lab/helpers/java.yml
@@ -29,6 +29,45 @@ scripts:
     - log: Showing GraalVM version
     - sh: $GRAALVM_HOME/bin/java -version
 
+  ensure-mandrel:
+    - sh: echo "${{config.jvm.mandrel.home:}}"
+      then:
+        - regex: ^$
+          then:
+            - sh: echo "${{config.jvm.mandrel.version:}}"
+              then:
+                - regex: ^$
+                  then:
+                    - log: No Mandrel version or home configured — skipping Mandrel setup
+                  else:
+                    # No local Mandrel path provided, use SDKMAN
+                    - script: sdk-select-java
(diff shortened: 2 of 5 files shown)


holly-cummins:
cc @zakkak

zakkak:
Hi @cdhermann, 

regarding "Support native builds with Mandrel next to GraalVM", I don't see the value in adding mandrel as an additional "runtime". Mandrel can already be used in place of GraalVM, so one just needs to set `graalvm-home`/`graalvm-version` to point to the desired Mandrel installation/version.

AFAIK there is no interest in being able to run the benchmark against different GraalVM distributions/versions by default. If so desired, one can get inspired by https://github.com/zakkak/plotting/tree/main/spring-quarkus-perf-comparison on how to do so.

So in summary I would say I am -1 on this @holly-cummins.

cdhermann:
Hello @zakkak,

> I don't see the value in adding mandrel as an additional "runtime"

If they behaved exactly the same, I would agree. However, my tests showed that even when using the same minor version, GraalVM CE and Mandrel differ in terms of which fields and methods are discovered. They also differ in image size (in my tests, Mandrel images were larger).

That said, you're most likely right that a performance comparison isn't the right place to highlight these kinds of differences.

edeandrea:
As @zakkak mentions you can run mandrel just by setting `--graalvm-version 25.0.2.r25-mandrel` rather than the default version `25.0.2-graalce` when calling `run-benchmarks.sh`.

I don't see the need to have mandrel be a separate runtime.

PR #113, opened 2026-01-05
Create hyperfoil yaml rather than running wrk directly

edeandrea:
This also produces the nice html reports, which are also downloaded by the `run-benchmarks.sh` script :

image

image

Fixes #42

PR #522, opened 2026-04-29
[OOTB] Create hyperfoil yaml rather than running wrk directly

### LINKED ISSUE #42: Create hyperfoil yaml rather than running wrk directly?
@franz1981 when we do the warmup & the test

https://github.com/quarkusio/spring-quarkus-perf-comparison/blob/d8a009dbe59b7f400ecf16d7cdf523968d85d353/scripts/perf-lab/main.yml#L365-L371

Would it be better to create a hyperfoil yaml file that has both stages and run `jbang run@hyperfoil` instead?

edeandrea:
## Summary
- Syncs PR #113 from `main` to `ootb`
- Replaces wrk-based load testing with Hyperfoil YAML configuration in perf-lab scripts and `stress.sh`
- Adds new `load-test-fixed-threads-read-all.hf.yml` Hyperfoil config file with configurable warmup, cooldown, and load test phases
- Parses Hyperfoil JSON output for throughput, connection errors, request timeouts, and HTTP 4xx/5xx error tracking
- Preserves all ootb-specific configuration (no OTEL monitoring, no `buildRequiresTestServices` changes)

Fixes #42

### DIFF
--- a/scripts/infra.sh
+++ b/scripts/infra.sh
@@ -136,7 +136,7 @@ else
 fi
 
 # Process the input options
-while getopts "c:dhm:p:rs" option; do
+while getopts "c:dhnm:p:rs" option; do
   case $option in
     c) CPUS=$OPTARG
        ;;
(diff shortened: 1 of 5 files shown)


edeandrea:
Since I'm not at home and dont have access to my linux machine(s) I'm going to do a test run in the lab before merging this.

PR #528, opened 2026-05-01
feat(micronaut): Adding a Micronaut example based on TCK published

jabrena:
- Added Micronaut example following TCK and OpenApi spec.
- Added ArchUnit test based on TCK.


edeandrea:
Hi @jabrena !

We wanted to thank you for this great work!

That being said, I don't think it is something that we will merge. We're happy to keep the PR open for anyone who might like to use it. Currently we are only focusing on spring vs quarkus and we don't want to take on the maintainance burden of a micronaut variant at the moment.

jabrena:
Hi, I undertand the argument about maintenance effort and I respect.

Note: As a User it was interesting to see the stats from all frameworks, not only Spring Boot & Quarkus.
https://github.com/quarkusio/spring-quarkus-perf-comparison/actions/runs/25223911302

Juan Antonio

edeandrea:
> Hi, I undertand the argument about maintenance effort and I respect.
> 
> Note: As a User it was interesting to see the stats from all frameworks, not only Spring Boot & Quarkus. https://github.com/quarkusio/spring-quarkus-perf-comparison/actions/runs/25223911302
> 
> Juan Antonio

Thanks for understanding! Its not just about a single variant. There would have to be a `jvm` (which is what this PR provides), `jvm-leyden`, `jvm-virtual-thread`, `jvm-virtual-thread-leyden`, & `native` variants of this, times 2 branches (`tuned` vs `ootb`). That's 10 total runtimes that we would need to maintain. It would also increase the run time in our lab by several hours per run, where it already takes over 8 hours to do a "full" run.

It just isn't something we want to do right now. Our primary focus is on Spring.

jabrena:
I understand it in a fully way.

I tried it :)

Cheers

Juan Antonio

PR #529, opened 2026-05-03
Set ParallelGC as the default garbage collector

### LINKED ISSUE #480: Consider switching to a lighter GC than G1
Running the benchmark for both Spring and Quarkus and collecting JFR events via --profiler syncjfr highlight a known alert

Image

This means that the available parallelism is not enough to make concurrent marking/refinement to benefit of it.
Translated: both the workload and availble cores buy nothing to G1 if not to slow it down in its concurrent phases.
Usually with so little cores and heap size, using parallel (or less, even serial?) is the way to go, and is what I usually recommend in production with this type of environment - not to mention that it can lower RSS to both frameworks.

Maybe is the case to use parallel for both and I believe I gave the same suggestion to @brunobat long time ago for a similar test, with some very very interesting results.

edeandrea:
## Summary
- Makes `-XX:+UseParallelGC` the default JVM argument across all benchmark scripts, documentation, and TCK
- Previously no GC was explicitly set, leaving the JVM default (G1GC on Java 21+)
- Updates `run-benchmarks.sh` default, `main.yml` qDup default, `stress.sh` hardcoded command, all example commands in docs, and adds ParallelGC as a mandatory setting in `TCK.md` Section 11

## Test plan
- [ ] Run `./run-benchmarks.sh -h` and confirm `-XX:+UseParallelGC` appears as the default for `--jvm-args`
- [ ] Verify `grep -r "UseParallelGC" scripts/ README.md TCK.md` shows all 12 expected occurrences

Fixes #480


edeandrea:
@holly-cummins the tuning label thing seemed to work fine (see https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/529#issuecomment-4366553834) but it looks like the sync attempted to run too (see https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/529#issuecomment-4366554192)

franz1981:
how the startup RSS was with it vs G1?

edeandrea:
With Parallel

| Scenario | Build Time (sec) | Startup Time (ms) | Startup RSS (MiB) | 1st Request RSS (MiB) | Load Throughput (tps) | Max Load RSS (MiB) |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| **spring4-jvm** | 7.14 | 10583.65 | 469.05 | 479.41 | 7479.40 | 673.44 |
| **spring4-virtual** | 7.11 | 10452.80 | 473.64 | 486.68 | 10055.27 | 817.09 |
| **quarkus3-jvm** | 13.92 | 4339.92 | 257.08 | 277.88 | 14714.55 | 555.94 |
| **quarkus3-virtual** | 13.90 | 4346.98 | 257.63 | 274.28 | 14965.32 | 527.87 |

With G1 (from https://github.com/quarkusio/benchmarks/blob/main/results/spring-quarkus-perf-comparison-latest-tuned.json)

| Scenario | Build Time (sec) | Startup Time (ms) | Startup RSS (MiB) | 1st Request RSS (MiB) | Load Throughput (tps) | Max Load RSS (MiB) |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| **spring4-jvm** | 6.96 | 10746.82 | 629.05 | 651.13 | 7615.56 | 816.68 |
| **spring4-virtual** | 7.03 | 10816.99 | 628.60 | 648.63 | 9179.09 | 801.58 |
| **quarkus3-jvm** | 14.06 | 4416.58 | 279.57 | 304.34 | 13264.69 | 779.62 |
| **quarkus3-virtual** | 14.05 | 4435.63 | 280.39 | 309.43 | 12240.92 | 741.83 |

Diff (parallel - G1)

> [!NOTE]
> This means that negative numbers mean ParallelGC was "better" than G1

| Scenario | Build Time Diff | Startup Time Diff | Startup RSS Diff | 1st Request RSS Diff | Throughput Diff |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **spring4-jvm** | +0.18s (+2.5%) | -163.17ms (-1.5%) | -160.00 MiB (-25.4%) | -171.72 MiB (-26.4%) | -136.16 tps (-1.8%) |
| **spring4-virtual** | +0.08s (+1.1%) | -364.19ms (-3.4%) | -154.96 MiB (-24.7%) | -161.95 MiB (-25.0%) | +876.18 tps (+9.5%) |
| **quarkus3-jvm** | -0.14s (-1.0%) | -76.66ms (-1.7%) | -22.49 MiB (-8.0%) | -26.46 MiB (-8.7%) | +1449.86 tps (+10.9%) |
| **quarkus3-virtual** | -0.15s (-1.1%) | -88.65ms (-2.0%) | -22.76 MiB (-8.1%) | -35.15 MiB (-11.4%) | +2724.40 tps (+22.3%) |

Gemini notes:

- The Throughput Surplus: Dataset 1 was significantly more "optimistic" regarding throughput, especially for the Quarkus 3 Virtual scenario, which performed 22.3% higher than the 3-iteration average seen in Dataset 2.
- The Memory Deficit: Dataset 1 recorded a much leaner memory footprint for Spring 4. Compared to the averages in the second set, Spring 4's startup RSS was ~25% lower in that initial run.
- Startup Stability: Startup times are the most consistent runtime metric here, with Dataset 1 only varying from the baseline by 1.5% to 3.4%.
- Build Inversion: While the absolute numbers are small, the build times for Spring were slightly higher in the first set, whereas Quarkus was slightly faster compared to the second set's average.

holly-cummins:
> @holly-cummins the tuning label thing seemed to work fine (see [#529 (comment)](https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/529#issuecomment-4366553834)) but it looks like the sync attempted to run too (see [#529 (comment)](https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/529#issuecomment-4366554192))

I just spotted that too and was (a) irritated and (b) mystified. Will investigate! 

I just checked https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/479, which also has the label, and it's sort of better and sort of worse. But mostly worse. It has the conflict comment, but it's fair enough, because I think the initial attempt was done before I added the label support. Then it comments about the tuning label, and it comments every time a push happens. I clearly should have used maintain-one-comment. I suspect it would be adding more conflict comments too, except that part of the script uses maintain-one-comment so it's sticking with the older one.

edeandrea:
@franz1981 @holly-cummins what do we want to do with this?

franz1981:
I would vote parallel everywhere really - ootb doesn't mean necessary "no settings" - because we set heap capacity still there

holly-cummins:
> I would vote parallel everywhere really - ootb doesn't mean necessary "no settings" - because we set heap capacity still there

I'm a strong -1 on this. IMO, ootb literally means "no settings". We set heap capacity to simulate a hardware constraint, but we shouldn't be adjusting tuning knobs to improve performance on a mode where we made a "we won't adjust tuning knobs to improve performance" promise. If we do, we need to rename it to `tuned`, and rename `tuned` to `more-tuned`. And then we might want to know what happens if we don't do tuning, so we'd need to make a new mode to see what happens with the default GC. We could call it 'ootb', maybe. And then we'd have to run 'ootb', 'tuned', and 'more-tuned'. See? If we avoid doing that, I just saved us increasing our combinatorics by a third! :) 

On the other hand, I'm very happy for us to merge. But we should document, to ensure the value of this change is broader than just a moved line on a chart. Who's going to write the blog post explaining the new recommendation? Something like "Choosing the right garbage collection mode for your Quarkus application"? I think that would be quite useful for people to have.

franz1981:
I hear the philosophical argument, but I think it proves too much. We already set `-Xms512m -Xmx512m` on ootb, which overrides two JVM ergonomic defaults: initial heap size (~1/64 of physical RAM by default) and max heap (~1/4 of physical RAM). Pinning Xms=Xmx is itself a well-known tuning practice to avoid resize pauses. So we're already turning tuning knobs — the line between "simulating constraints" and "tuning" isn't as clean as it sounds.

More concretely: the JVM selects G1GC based on the host environment (CPUs ≥ 2, physical memory ≥ 1792MB) — it doesn't look at `-Xmx` for that decision. So by overriding heap to 512m but leaving the GC default, we've created a mismatch: small-heap resources paired with a GC designed for large heaps. That's not really what any user gets "out of the box" — it's an artificial worst case that doesn't correspond to any real deployment.

That said, I'm fine merging this as-is for tuned and having the ootb discussion separately if needed.

Re: blog post — agree, that's a good idea regardless.

edeandrea:
> > I would vote parallel everywhere really - ootb doesn't mean necessary "no settings" - because we set heap capacity still there
> 
> I'm a strong -1 on this. IMO, ootb literally means "no settings". We set heap capacity to simulate a hardware constraint, but we shouldn't be adjusting tuning knobs to improve performance on a mode where we made a "we won't adjust tuning knobs to improve performance" promise. If we do, we need to rename it to `tuned`, and rename `tuned` to `more-tuned`. And then we might want to know what happens if we don't do tuning, so we'd need to make a new mode to see what happens with the default GC. We could call it 'ootb', maybe. And then we'd have to run 'ootb', 'tuned', and 'more-tuned'. See? If we avoid doing that, I just saved us increasing our combinatorics by a third! :)
> 
> On the other hand, I'm very happy for us to merge. But we should document, to ensure the value of this change is broader than just a moved line on a chart. Who's going to write the blog post explaining the new recommendation? Something like "Choosing the right garbage collection mode for your Quarkus application"? I think that would be quite useful for people to have.

I thought we agreed this change is only on `main`? I thought that conversation was settled.

franz1981:
@edeandrea 

> I thought we agreed this change is only on main? I thought that conversation was settled.

I know, but in these days I've quickly checked that the Red Hat UBI OpenJDK base images that Quarkus uses for container builds already default to `-XX:+UseParallelGC` ([source](https://github.com/jboss-container-images/openjdk/blob/ubi9/modules/jvm/artifacts/opt/jboss/container/java/jvm/java-default-options#L50)). 
G1GC in our ootb benchmark is less representative of real ootb than ParallelGC.

edeandrea:
I would think which garbage collector to use would be application-dependent :)

franz1981:
> I would think which garbage collector to use would be application-dependent :)

Sure, it's a mix of app and environment — but that's exactly the point. 
In the environment we're simulating (containerized, constrained memory), the platform default is ParallelGC - and what most would choose. 
We're not choosing a GC to favor an app.

The fact that Spring benefit so much more of it is exactly because of been a CPU intensive test (an env where the CPU time is scarse, for a mix of reasons) and because Spring need it more than Quarkus, being heavier.

The "inverted parallelism" hint in the issue I have opened re it, is a sign of this

edeandrea:
Can we come to a consensus? I'd like to cross this off.

Right now I'm only planning on this as the default for `tuned`. More and more I feel like `ootb` is more like `baseline`, and anything in `tuned` is to improve over the baseline.

holly-cummins:
+1, let's merge this. We can revisit ootb later if we need to,

franz1981:
@edeandrea it's fine for me too!

hatahet:
Are the latency percentiles available for both GCs?

PR #554, opened 2026-05-13
Add extra help for plotting results

holly-cummins:
I was going to add a claude skill for plotting results, and then I realised a script might work as well, and also be useful for humans.


edeandrea (approved):

edeandrea:
@holly-cummins it says this was synchronized to the `ootb` branch, but I don't see a PR for it on the `ootb` branch...

PR #556, opened 2026-05-15
Add load test for fixed-rate sustained load testing

holly-cummins:
See dicussion here: https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/473#discussion_r3189765045

This is phase 1 of #474, to add a test which gives meaningful energy results. Then #474 can be rebased onto this, to add the energy metrics. 

- Adds a new `soak-test` that runs Hyperfoil at a configurable fixed request rate (`constantRate` phase) for an extended duration, measuring sustained throughput stability
- New Hyperfoil YAML benchmark file (`soak-test-fixed-rate-read-all.hf.yml`) with parameterized rate, duration, warmup, and connection settings
- CLI options `--soak-rate` and `--soak-duration` in `run-benchmarks.sh` to configure the test

Claude's run these changes locally, and claims they work, but the testing seems a bit half-hearted to me, so I'm going to check further. :) 

- [x] Verified `run-benchmarks.sh --help` displays new soak options correctly
- [x] Validated `--soak-rate` and `--soak-duration` reject non-integer input
- [x] Confirmed CLI options pass through to qDup state correctly
- [x] Validated Hyperfoil YAML parses and runs (connection refused expected without app, but template loads and phases execute)

---

After a bit of coaxing, Claude says, "⏺ Soak test ran successfully against the live app - 966 requests at ~97 tps (close to the configured 100 tps), 0 errors, mean latency 1.54ms. "

cc [diegolovison](https://github.com/diegolovison)
[l](https://github.com/quarkusio/spring-quarkus-perf-comparison/pull/473#discussion_r3189765045)


franz1981:
This is a sustained fixed-rate load test, not a soak test (soak implies hours/days to find leaks and resource exhaustion).

A few things to fix:

**1. `maxSessions` must be much larger than `connections`.** Both are set to `CONNECTIONS` (100). With `constantRate`, sessions are created at the target rate even when all connections are busy — they queue up waiting. If `maxSessions` equals `connections`, Hyperfoil can't create the queued sessions, so blocked time is never recorded. The overload becomes invisible — worse than [coordinated omission](https://www.scylladb.com/2021/04/22/on-coordinated-omission/) because the evidence is never collected. Size it for the longest disruption you want to observe: `maxSessions = rate × max_expected_delay` (e.g., 5000 × 2s GC pause = 10,000).

**2. Size `sharedConnections` with headroom.** With HTTP/1.1, each connection handles one request at a time. Max throughput = `connections / avg_latency`. At 100 connections and 20ms latency that's exactly 5000 req/s — every connection must be busy 100% of the time. The average rate is constant, but Hyperfoil catches up if it falls behind — if the first half of a second only managed 2000 requests (client-side scheduling delay, GC), the second half fires 8000 to compensate. That catch-up burst needs connections available right now. On the server side, individual response times vary — a single slow response (GC pause, slow query) holds a connection longer, reducing available capacity for the burst. Set connections generously (500-1000) so there's always slack to absorb both catch-up bursts and server-side latency variance. Idle connections are (relatively) cheap.

**3. `variance: false` on `constantRate` phases.** Disables Poisson inter-arrival times. More reproducible for CI comparisons.

**4. Use `increasingRate` for warmup.** A `constantRate` warmup hits full rate from t=0, forcing the server to oversize thread/connection pools for the initial burst. An `increasingRate` ramp lets pools stabilize at the size the target rate actually needs, so the soakTest measures steady-state behavior, not burst-recovery.

**5. Add `failurePolicy: CONTINUE`.** Hyperfoil's default SLA fails on any connection blocking (`blockedRatio: 0`) and the default `CANCEL` policy aborts all remaining phases. Your warmup saturates the pool → soakTest gets cancelled → you get warmup stats only with no indication why. `failurePolicy: CONTINUE` lets the test complete and reports SLA failures in the output instead. Being fixed for wrk/wrk2 in [Hyperfoil#793](https://github.com/Hyperfoil/Hyperfoil/pull/793), but user-defined benchmarks need this in the YAML for now.

**6. Use Hyperfoil's feedback to tune the rate.** You can't know the right `SOAK_RATE` upfront. With `failurePolicy: CONTINUE`, Hyperfoil always completes and reports what went wrong — blocked connections, session exhaustion, growing latency. Start with generous sessions/connections, pick a rate below both SUTs' max throughput, run, read the feedback, adjust, repeat. When the stats are clean, compare latency distributions between Spring and Quarkus at that rate — the framework with lower tail latency is the better performer.

holly-cummins:
Thanks, @franz1981! I think I've addressed these and pushed:

  1. maxSessions >> connections: Set maxSessions to 10,000 (default), sharedConnections to 500 — sessions can queue when connections are busy, so   
  blocked time is actually recorded                                                                                                                 
  2. sharedConnections with headroom: 500 instead of 100 — gives slack for catch-up bursts and server-side latency variance                         
  3. variance: false: Disables Poisson inter-arrival for reproducible CI results                                                                    
  4. increasingRate warmup: Ramps from 1 to target rate instead of hitting full rate at t=0, letting server pools stabilize                         
  5. failurePolicy: CONTINUE: Test completes and reports SLA failures instead of aborting remaining phases                                          
  6. Renamed params: SOAK_* → LOAD_* and phase soakTest → loadTest since this is a sustained fixed-rate load test, not a soak test

scripts/perf-lab/main.yml
```diff
@@ -58,7 +62,7 @@ states:
   PAUSE_TIME: 5
   PROFILER_JVM_ARGS:
   BASE_JAVA_CMD: ${{APP_CMD_PREFIX}} java ${{config.jvm.memory}} ${{config.jvm.args}} ${{PROFILER_JVM_ARGS}}
-  TESTS : [measure-time-to-first-request, measure-rss, run-load-test]
+  TESTS : [measure-time-to-first-request, measure-rss, run-load-test, soak-test]
```
franz1981: Soak test

franz1981 (commented):
What kind of metrics we are interested in for this type of benchmark? I would say we want to expose the latency profile (p50, pXY,..p99, max) once we are at steady state, but I am not sure is a good idea to average percentiles. I have to think about it

franz1981:
The core problem with averaging percentiles Averaging p99 across iterations isn't statistically meaningful — the p99 of the merged population won't generally match the average of per-run p99s. The two correct approaches are:

1. Merge per-iteration HDR histograms and compute percentiles from the combined population — Hyperfoil should records these in all.json and they can be merged correctly. If we don't we can add this option in hyperfoil easily (iirc Gunnar asked the same some time ago, probably for the same reason)
2. Report the distribution of per-iteration p99s — median, max, and coefficient of variation — which tells you both the typical tail latency and how stable it is across runs.


**On throughput**: since the fixed TPS is deliberately chosen to be sustainable for both frameworks, collecting average throughput doesn't add much comparative signal — it'll be near-identical by design. It's worth keeping as a sanity check (achieved rate vs. target rate) but shouldn't be the headline metric.

What would be most useful instead:

- Latency tails from merged histograms (p50, p99, p99.9) — the statistically correct population-level view
- Per-iteration p99 distribution (median + max + CV) — reveals stability, not just central tendency
- Blocked count / blocked ratio — the key overload signal at fixed rate; if a runtime struggles to sustain the rate, connections back up and this shows it, whereas throughput (nearly at target) would hide it

holly-cummins:
> What kind of metrics we are interested in for this type of benchmark? I would say we want to expose the latency profile (p50, pXY,..p99, max) once we are at steady state, but I am not sure is a good idea to average percentiles. I have to think about it

The metric I personally care about is energy consumption. :) 

I agree that throughput is unlikely to be interesting, and the percentiles need careful thought (but might be interesting if treated appropriately). 

I created this PR to unblock the energy PR, so I don't want to get *too* caught up on the details of the load test metrics, when the most interesting metric hasn't been added yet.

scripts/perf-lab/main.yml
holly-cummins: Good spot, fixed!

franz1981:
So all good from me, but please run it once in the perf lab without updating the Horreum results so we can read the extracted data and check if is fine for all the configurations @holly-cummins

scripts/perf-lab/run-benchmarks.sh
```diff
@@ -332,6 +340,8 @@ if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
   NATIVE_QUARKUS_BUILD_OPTIONS=""
   NATIVE_SPRING3_BUILD_OPTIONS=""
   NATIVE_SPRING4_BUILD_OPTIONS=""
+  SOAK_DURATION=150
+  SOAK_RATE=5000
```
holly-cummins: @franz1981 points out this is too high for Spring native, so adjusting it down.

PR #588, opened 2026-06-01
Use quarkus theme for architecture diagram

holly-cummins:
Now that we have a shared quarkus theme for plantuml, we can start using it for our diagrams. My first attempt was strictly themed, and that was rather grey, so I reverted to having some inline styling, but more Quarkus-branded:
image


holly-cummins:
> 🤖 Sync to `ootb` was skipped — all commits had conflicts and could not be cherry-picked. This PR may need to be manually applied to the `ootb` branch, or it may not be relevant to that branch. 🔄

Hmm, odd that there'd be conflicts on this. I wonder if something needs to be manually sorted out.

edeandrea (approved):

PR #601, opened 2026-06-09
Fix async-profiler to capture load test instead of warmup phase

### LINKED ISSUE #600: async-profiler captures warmup instead of load test since Hyperfoil migration
Since commit a6e403e ("Create hyperfoil yaml rather than running wrk directly", #113), the async-profiler JFR profiling captures the **warmup phase** instead of the **load test (measurement) phase**.

### Root cause

Before the Hyperfoil migration, the qDup sequence was:

```
wrk warmup 2m → sleep 30s → signal: LOAD_STEADY_STATE_START → wrk load test 30s
```

The `async-profiler-run` script waits for `LOAD_STEADY_STATE_START` + 10s delay, then profiles for 10s. With wrk, this captured seconds 10-20 of the **load test**.

After the migration, the sequence is:

```
signal: LOAD_STEADY_STATE_START → Hyperfoil starts (warmup 2m + cooldown 30s + load test 30s)
```

The signal now fires **before Hyperfoil warmup begins**. The profiler captures seconds 10-20 of the **warmup phase** — not the load test.

### Impact

All async-profiler JFR data collected since the Hyperfoil migration (commit a6e403e) is from the warmup phase, making `jfrconv --classify` tier breakdowns reflect warmup compilation state rather than steady-state performance.

### Fix

Move `signal: LOAD_STEADY_STATE_START` to fire after the Hyperfoil warmup+cooldown completes and before the load test begins. This requires either:
1. Using Hyperfoil's `logStart` phase (already present in the YAML) to signal when the load test starts
2. Or computing the delay based on warmup duration + cooldown duration (150s) and adding it to the profiler delay

edeandrea:
## Summary

- Since the Hyperfoil migration (commit a6e403e), async-profiler JFR/flamegraph captures the **warmup phase** instead of the **load test phase** because `LOAD_STEADY_STATE_START` fires immediately after app startup — before Hyperfoil even begins its 2-minute warmup
- Adds a dedicated `watch-hyperfoil-log` script that watches Hyperfoil's internal log (`hyperfoil.local.log`) for the `loadTest` phase transition to `RUNNING`, then fires the signal at the correct time
- Reduces the async-profiler delay from 10s to 5s since the signal now fires when the load test actually starts

## Changes

- **Removed** the premature `signal: LOAD_STEADY_STATE_START` that fired right after app startup
- **Added** `watch-hyperfoil-log` script: watches `hyperfoil.local.log` for `loadTest.*changing status.*RUNNING` with a 5m timer and `tail -F` (handles file not yet existing during Hyperfoil startup)
- **Added** async invocation of `watch-hyperfoil-log` after Hyperfoil is launched
- **Reduced** async-profiler delay from 10s to 5s
- **Fixed** `watch-log` abort message to match the actual 1m timer duration

## Test plan

- [x] Run benchmark with `--profiler flamegraph` and verify the run completes without the `watch-hyperfoil-log` aborting
- [ ] Verify async-profiler output captures data from the load test window (after warmup+cooldown), not the warmup phase
- [ ] Run benchmark without profiling to confirm no regressions

Fixes #600

### DIFF
--- a/scripts/perf-lab/main.yml
+++ b/scripts/perf-lab/main.yml
@@ -625,7 +625,7 @@ scripts:
                     suffix: ${{PROFILER_LOGFILE}}
                     format: ${{config.profiler.name}}
                     events: ${{config.profiler.events}}
-                    delay: 10s
+                    delay: 5s
         - script:
             name: watch-log
             async: true
@@ -633,7 +633,6 @@ scripts:
             log_file: ${{LOG_FILE}}
             log_regex: ${{APP_LOG_REGEX}}
         - wait-for: LOG_REGEX_REACHED
-        - signal: LOAD_STEADY_STATE_START
         - sh: |
             ${{LOAD_GEN_CMD_PREFIX}} jbang \
               -Dio.hyperfoil.rootdir=${{HYPERFOIL_LOGS_DIR}}/${{RUNTIME_NAME}}-${{ITERATION}} \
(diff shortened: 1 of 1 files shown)


edeandrea:
@franz1981 does this look ok to you?

PR #660, opened 2026-07-10
Migration result

aureamunoz:
### Setup
- **Skill**: `migrate-spring-to-quarkus` (from [quarkusio/quarkus-skills](https://github.com/quarkusio/quarkus-skills)) executed against `springboot3/` with "Spring compatibility" strategy. The quarkus3-spring-compatibility/ has been overrided with the migration result so the diff shows exactly what the skill changed compared to the script-generated baseline.
- **Skill result**: compiles, starts on localhost:8080, tests pass

### Java source: IDENTICAL

Zero differences in `src/main/java/`. The skill produced exactly the same code as the script.

### Differences

| Aspect                            | Script (`quarkus3-spring-compatibility`)    | Skill (`springboot3`)                            | Verdict                                                   |
| --------------------------------- | ------------------------------------------- | ------------------------------------------------ | --------------------------------------------------------- |
| **artifactId/name**               | `quarkus3-spring-compatibility`             | `springboot3` (kept original)                    | Neutral — script uses a pre-written pom with its own name |
| **finalName**                     | `app`                                       | `springboot3`                                    | Neutral — repo convention                                 |
| **pom description**               | "using Spring compatibility layer"          | "migrated from Spring Boot 3"                    | Neutral                                                   |
| **`quarkus-container-image-jib`** | Included                                    | Not included                                     | **Skill gap** — see below                                 |
| **SQL data file**                 | `data.sql` (Spring convention)              | `import.sql` (Quarkus convention)                | **Skill better** — `import.sql` is the Quarkus convention |
| **Tests**                         | Deletes ALL, copies `e2e/` from `quarkus3/` | Migrates originals to `@QuarkusTest`             | **Skill better** — preserves original test coverage       |
| **`SpringBoot3ApplicationTests`** | Deleted                                     | Migrated to `@QuarkusTest` with `contextLoads()` | Skill preserved it                                        |
| **application.yml**               | Extra comment + trailing newline            | Same functional content                          | Irrelevant                                                |

### About `quarkus-container-image-jib`

The skill did not include `quarkus-container-image-jib` in the migrated pom. In this repo, **all Quarkus modules** have it (`quarkus3/`, `quarkus3-virtual/`, `quarkus3-spring-compatibility/`) because the perf lab scripts deploy apps as containers.

This is **not a general skill gap** — the skill shouldn't add infrastructure extensions that weren't in the original Spring project (Spring Boot uses `spring-boot-maven-plugin` for executable JARs, not container images directly). But for this specific repo, the migrated pom would need `quarkus-container-image-jib` to be equivalent to the other Quarkus modules.

### Conclusion

The skill produces a functionally equivalent result to the script, with two improvements (tests migrated instead of deleted, correct SQL filename) and one minor gap (container image extension that is repo-specific, not migration-related).

This PR is not intended to be merged, it's more for allowing discussion and share the result of an AI skill-driven migration and reflect on the differences with the current bash script approach.


quarkus3-spring-compatibility/src/main/resources/application.yml
```diff
@@ -7,7 +7,6 @@ quarkus:
   rest:
     jackson:
       optimization:
-        # See https://quarkus.io/blog/quarkus-metaprogramming/
```
holly-cummins, line 10: Can we keep this?

holly-cummins:
We need to think carefully about this, because one of the things that is half-completed on my laptop is an automated refresh of the compatibility project, to make sure it stays in sync with the main quarkus and spring projects. It will run in CI, so will use the bash scripts. It would be annoying if we get a lot of churn back and forth while the bash script and migration fight over things like line endings and file names. So we maybe also want to adjust the bash script for some of these things (but we don't want to adjust it too much, because people need to be able to live code every step of what's in the script in talks).

aureamunoz:
> We need to think carefully about this, because one of the things that is half-completed on my laptop is an automated refresh of the compatibility project, to make sure it stays in sync with the main quarkus and spring projects. It will run in CI, so will use the bash scripts. It would be annoying if we get a lot of churn back and forth while the bash script and migration fight over things like line endings and file names. So we maybe also want to adjust the bash script for some of these things (but we don't want to adjust it too much, because people need to be able to live code every step of what's in the script in talks).

Agreed, there's no risk of churn here atm, the skill can't run in CI due to AI licensing, so only the bash script would drive the automated refresh. This PR is purely a comparison snapshot. That said, if any of the differences make sense independently (e.g. renaming data.sql to import.sql to follow Quarkus convention), those could be adopted in the bash script on their own merits.

PR #661, opened 2026-07-11
Refactor: modularize profilers into a plugin-like architecture

diegolovison:
This PR is still a draft because it requires @franz1981 review.
- I already tested locally.

### **Motivation**

The core `main.yml` pipeline was becoming cluttered and difficult to maintain due to a monolithic `if/else` regex block handling profiler selection. Adding new profiling tools required modifying the core pipeline directly, increasing complexity and the risk of introducing errors.

### **The New Feature**

* **Modular Architecture:** Extracted all profiler logic out of `main.yml` and reorganized it into a plugin-based architecture.

* **Dedicated Configurations:** Created isolated configuration files for each profiler (`profiler-flamegraph.yml`, `profiler-jfr.yml`, `profiler-syncjfr.yml`, `profiler-none.yml`, and `profiler-perf-stat.yml`) located in a new `./profilers/` directory.

* **`perf stat` Support:** Added a new profiler option to capture hardware performance counter statistics using Linux `perf stat`.

### **Why This is Better Than Before**

* **Cleaner Core Workflow:** `main.yml` no longer cares about *which* profiler is running; it simply executes dynamic hooks (`${{config.profiler.name}}-install` and `-run`), keeping the main script drastically shorter and easier to read.

* **Highly Scalable:** Adding new profilers in the future is now trivial. You only need to drop a new YAML file into the `./profilers/` folder and update the bash script options, completely avoiding changes to the core workflow.

* **Separation of Concerns:** Each profiler now entirely manages its own dependencies, wait states, and custom run commands (like `syncjfr` injecting JVM flags) natively within its own file.

PR #681, opened 2026-07-21
Switch warmup to multi-cycle: 10 x (15s load + 15s cooldown)

### LINKED ISSUE #633: Move to a different warmup to help Spring being properly hot
While working at #591 and written https://quarkus.io/blog/when-the-jit-cant-keep-up/ I've noticed that the perf-lab CFS OS scheduler is making Spring much much slower to warmup compared to new Linux versions using the new OS scheduler e.g. latest ubuntu. 

Users which run our benchmarks on bare-metal GoogleCloud or AWS with it would surely see a much different picture, see

perf lab:

Image

newer machine w fedora 43:

Image

The fact we average the tps for the load throughput has hidden that Spring was still warming up (more on this on the article itself).

I've decided to move to a 10 iterations including shorter cooldown + loads, before the actual load (total 5 mins 30 seconds for each framework) to make sure each framework is properly warmed up, getting instead

Image

and C2 compile threads activity

Image

which shows that everything is hot and the activity very little at the 10th iteration (no more green or pink), before the load.

But still, despite Spring4 jvm is warmed up, its tps over time is still very unstable (better than before the 10-iterations change, but still...) and it looks to be OS-dependent:
- spring4-jvm relies on Hicari CP which uses `Thread::yield` heavily and cause storm of non voluntary context switches causing the unstable tps (let's remember we have +100 active platform threads!)
- quarkus instead uses Agroal which doesn't have such problem
- spring4-loom `Thread::yield` is no longer using OS's `yield` but performing a `lighter` context switch to the Loom scheduler (ForkJoin pool) which makes it more stable and behave the same across different OSs versions

Some of the analysis I've performed are reported https://gist.github.com/franz1981/3db7a9c862d41786864950959a5d79fb#scheduling-analysis-why-loom-helps-spring-but-not-quarkus-100-connections-4-cores which explain the HicariCP behaviour.

TLDR:
- it's better to use a n-iterations approach (as https://github.com/franz1981/spring-quarkus-perf-comparison/tree/warmup-iterations does) because it improve the chances Spring is fully warmed up
- we still "fail" to have fully stable tps for Spring4-jvm due to perf lab OS scheduler (to be fair, is not so strange to run on RHEL < 10...), but we cannot really do much about it as it depends by HicariCP
- computing tps with averages clearly doesn't help: maybe we should instead report throughput + variance (@holly-cummins using an error bar, visually)? If we collect avg + variance over N indipendent runs, we should probably use t-student with some confidence interval to report it (@edeandrea ): this will help in case both Quarkus or Spring won't achieve "stable enough" tps and will highlight the benefits of using other variants (e.g. loom), wdyt?

And if you're asking, why Franz is spending so many cycles to make Spring appear better?
It's because of fairness, although the fact that spring is so heavy to warmup (in general) is a given fact, but we want any users which run on modern/old linux properly sized machine to have similar ratio/xp, but as usual, YMMV!

franz1981:
On CPU-bound workloads with many platform threads (e.g., Spring Boot with 100 connections on 4 cores), C2 compiler threads are starved during a single continuous warmup. The application never reaches peak performance before the measurement phase begins.

Multi-cycle warmup alternates short load bursts with cooldown gaps. Each cooldown gives C2 compiler threads CPU time to drain the compilation queue. After 10 cycles (5 minutes total), hot methods are fully compiled and the load test measures peak throughput.

Validated on the perf lab: spring4-jvm goes from ~7,850 TPS (old warmup) to ~8,500 TPS (multi-cycle). Quarkus reaches peak by cycle 2-3.

Uses Hyperfoil maxIterations for compact YAML. All parameters have configurable defaults via !param (WARMUP_CYCLES, WARMUP_DURATION, WARMUP_PAUSE_DURATION). No changes to run-benchmarks.sh or Jenkins config needed.

Also increases the watch-hyperfoil-log timeout from 5m to 10m to accommodate the longer warmup.

Fixes #633

### DIFF
--- a/scripts/perf-lab/load-tests/load-test-fixed-threads-read-all.hf.yml
+++ b/scripts/perf-lab/load-tests/load-test-fixed-threads-read-all.hf.yml
@@ -3,19 +3,28 @@
 # =============================================================================
 #
 # This configuration file defines a load test scenario for the application's read-all
-# endpoint. It includes a warmup phase followed by a delay, then the main load test phase.
+# endpoint. It uses multiple warmup+cooldown cycles to ensure the C2 compiler has
+# enough CPU time to finish compiling hot methods before the measurement phase.
 #
-# All phases target the same HTTP endpoint.
+# Each warmup burst exercises the code paths under full closed-loop load, and each
+# cooldown gives the C2 compiler threads CPU time to drain the compilation queue.
+# After all cycles complete, the load test measures peak throughput.
 #
-# This is to replace how we did things exactly as before:
-# warmup
-# wrk -t 2 -c 100 -d 2m --timeout 2m http://localhost:8080/fruits
+# Why multi-cycle warmup:
+#   On CPU-bound workloads with many platform threads (e.g., 100 connections on
+#   4 cores), C2 compiler threads are starved during a single continuous warmup.
+#   The application runs suboptimal C1-compiled code for the entire warmup and
+#   never reaches peak performance before measurement begins. Multi-cycle warmup
+#   gives C2 periodic CPU time during cooldown gaps to compile hot methods.
 #
-# cooldown
-# sleep 30s
-#
-# loadTest
-# wrk -t 2 -c 100 -d 30s --timeout 30s http://localhost:8080/fruits
+# Breaking changes from the previous single-warmup format:
+#   - Total warmup time increases from ~2.5min to ~5.5min per runtime
+#   - Hyperfoil series CSVs use warmup_NNN naming (e.g., warmup_000.getAllFruits.3.series.csv)
+#     instead of warmup.getAllFruits.2.series.csv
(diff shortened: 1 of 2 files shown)


franz1981:
It doesn't change metrics.json but we can manually download the hyperfoil html reports and manually verify that the tps is stable. In a future, we have to parse the csv files from hyperfoil and compute what we want e.g. error bars, violin distribution etc etc

@holly-cummins 

@diegolovison check if it interacts with what you have done on #661 :pray: 

@edeandrea I haven't yet run it outside locally. It shouldn't require any changes in our perf lab jenkins file unless we want to customize something. 
Currently we were not exposing in the sh script ANY of the warmup/cooldown configs, so I didn't done it here as well.
It can be a follow up pr if we feel to be worthy

franz1981:
The PR per se is done, hope today to fire a CI job to see if it works as intended across all tests

